PR #44, opened 2015-05-30
Added the extensions option to the CPPLINT.cfg option file

massenz:
Currently the CPPLINT.cfg config file does not accept the `extensions` option.
This is limiting, for example, in the `CppStyle` Eclipse plugin, which does not allow passing command-line arguments.

Adding the functionality is trivial.

Also, this patch adds the extensions to the `_valid_extensions` set, it does not replace it.

Tested with a .hpp file and the following CPPLINT.cfg:

```
$ cat ../mesos/CPPLINT.cfg 
extensions=hpp
```

and then:

```
$ cpplint.py src/common/http.hpp 
src/common/http.hpp:24:  Found C system header after C++ system header. Should be: http.h, c system, c++ system, other.  [build/include_order] [4]
src/common/http.hpp:26:  Found C system header after C++ system header. Should be: http.h, c system, ...
Done processing src/common/http.hpp
```

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -1666,7 +1666,7 @@ def GetHeaderGuardCPPVariable(filename):
   filename = re.sub(r'/\.flymake/([^/]*)$', r'/\1', filename)
   # Replace 'c++' with 'cpp'.
   filename = filename.replace('C++', 'cpp').replace('c++', 'cpp')
-  
+
   fileinfo = FileInfo(filename)
   file_path_from_root = fileinfo.RepositoryName()
   if _root:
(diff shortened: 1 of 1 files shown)


mplewis:
I'd really like this functionality too. Can someone please review this pull request?

massenz:
This repo looks "abandoned" to me - nobody looks at PRs: this has been sitting here for months now.
If you want this, I've forked it and added a bunch of other functionality - see in the `master` branch:
https://github.com/massenz/styleguide/tree/master

If you want more functionality, please file a request there: I DO look at PRs :)

tkruse:
Hi @massenz, I had extracted cpplint into an own github repo back when the google styleguide was in googlecode svn. Here: https://github.com/tkruse/cpplint
I have additionally made it available at pypi:
https://pypi.python.org/pypi/cpplint

Would you be interested in having your changes to cpplint be included there?

tkruse:
See https://github.com/tkruse/cpplint/tree/add_extensions

PR #45, opened 2015-06-01
Added an extensible set() for header extensions

massenz:
Currently, `styleguide` only supports `.h` header files; this patch enables a more flexible approach, where any of a set of extensions (initialized, by default to be only `.h`, so no change from current behavior) will be treated as header files.

This is, for example, an issue for [Apache Mesos](http://mesos.apache.org) where we have "hacked" `cpplint.py` to include `.hpp` too.

This approach is cleaner and extensible.
Once (if?) you guys accept my other patch for `extensions` I will add support for `header_extensions` to `CCPLINT.cfg` too (if I do it here, it'll cause "merge confusion").

NOTE - the patch is not entirely PEP8 compliant: I tried instead to be consistent with existing code.

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -1797,9 +1797,9 @@ def CheckHeaderFileIncluded(filename, include_state, error):
   for ext in _header_extensions:
       headerfile = filename[0:len(filename) - 2] + ext
       if not os.path.exists(headerfile):
-        return
+        continue
       headername = FileInfo(headerfile).RepositoryName()
-      first_include = 0
+      first_include = None
       for section_list in include_state.include_list:
         for f in section_list:
           if headername in f[0] or f[0] in headername:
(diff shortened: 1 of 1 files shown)


cpplint/cpplint.py
```diff
@@ -1780,21 +1794,22 @@ def CheckHeaderFileIncluded(filename, include_state, error):
...
-  if not os.path.exists(headerfile):
-    return
-  headername = FileInfo(headerfile).RepositoryName()
-  first_include = 0
-  for section_list in include_state.include_list:
-    for f in section_list:
-      if headername in f[0] or f[0] in headername:
+  for ext in _header_extensions:
+      headerfile = filename[0:len(filename) - 2] + ext
+      if not os.path.exists(headerfile):
```
solokacher, line 1799: return if any of the header file does not exist?

massenz, line 1799: great catch!
replaced with `continue`

PR #48, opened 2015-06-08
Ignore Doxygen Long Lines

mattyclarkson:
There is a special command in Doxygen that allow documentation to be copied from another function but the full declaration must be put on one line. This result in an error in `cpplint` but to solve the error `// NOLINT` must be put on the line, which means that Doxygen fails to pick up the overload. This ignores any single line Doxygen documentation copy commands:

```
/// @copydoc SomeOverloadFunction(const int lots, const int of, const int lengthy, const int parameters)
```


tkruse:
See https://github.com/tkruse/cpplint/tree/doxygen

vapier:

PR #57, opened 2015-07-25
Fix running cpplint.py from the Windows command line

mbikovitsky:
On Windows, if the Python interpreter in your path is Python 3, executing cpplint.py from the command line will fail but nothing will be output to the console.

This little fix allows py.exe (the Python Launcher for Windows) to correctly choose the right interpreter.


vapier:

PR #70, opened 2015-09-21
Add "typeid" in the exclusion list in CheckCStyleCast"

akira-okumura:
Using C++ `typeid` function (e.g., `typeid(int) == typeid(bool)`) generates the following error.

> generates an error "All parameters should be named in a function [readability/function] [3]"

Please add `typeid` in the exclusion list in `CheckCStyleCast`.


vapier:

PR #71, opened 2015-09-23
Fixed cppguide.xml redirect.

DavidLemayian:
The [C++ formatting rules url](https://google.github.io/styleguide/cppguide.xml#Formatting) linked [here](https://google.github.io/styleguide/javascriptguide.xml#Code_formatting) throws the following error:

screen shot 2015-09-23 at 10 00 13 pm

This PR fixes the error by adding a redirect template that receives a url to redirect to and a
page title. It also keeps the same best practice of using a BODY onLoad redirect, with a meta-refresh fallback.

### DIFF
--- a/cppguide.xml
+++ b/cppguide.xml
@@ -1,18 +1,4 @@
-<!DOCTYPE html>
-<html>
-<head>
-  <meta charset="utf8">
-  <meta http-equiv="content-type" content="text/html;charset=utf-8">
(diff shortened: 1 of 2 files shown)


tonyruscoe:
I stumbled across this PR while searching for code to create a working meta refresh redirect in XML/XSL, because I’m going to need something similar when I migrate the HTML/CSS Style Guide from XML to HTML.

Is anyone able to merge this?

PR #72, opened 2015-10-01
Single line lambdas

mattyclarkson:
Allows the following code to be allowed:

```
const auto lambda = [](const int i) { return i; };
const auto result = std::any_of(vector.begin(), vector.end(), [](const int i) { return i > 0; });
```

Which is useful because `clang-format` collapses simple lambdas.


vapier:

PR #74, opened 2015-10-02
Explicit variadic constructors

mattyclarkson:
Makes sure that variadic constructors are marked explicit. We want to avoid this:

```
#include <iostream>
#include <vector>

struct A {
    template<typename... Args>
    explicit A(const int one, Args&&... args)
        : vector_({one, std::forward<Args>(args)...}) {}
    std::vector<int> vector_;
};

std::ostream& operator<<(std::ostream& stream, const A& a) {
    char comma = '[';
    for (auto&& element : a.vector_) {
        stream << comma << element;
        comma = ',';
    }
    if (a.vector_.empty()) stream.put(comma);
    return stream.put(']');
}

static void print(const A& a) {
    std::cout << a << std::endl;
}

int main()
{
    A a(1);
    print(a);
//    print(2); <-- Don't want implicit conversion!!
    return 0;
}
```

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -2694,6 +2694,7 @@ def CheckForNonStandardConstructs(filename, clean_lines, linenum,
       constructor_args[i] = constructor_arg
       i += 1
 
+    variadic_args = [arg for arg in constructor_args if '&&...' in arg]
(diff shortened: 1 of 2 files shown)


vapier:

Issue #36, opened 2015-05-20
styleguide maintenance


GoogleCodeExporter:
Hi,

I see some problem with the maintenance of the styleguide as a project:

- The project contains no README, LICENSE, CONTRIBUTING or CHANGELOG file (the project homepage contains that information instead, of course, but that's not how to do open-source, forks should have this information)
- The project homepage does not link to this issue tracker
- The homepage does not mention a maintainer, it is not clear at all who will respond to issues, and who may decide about style changes.
- some HTML sources have xml file ending
- Eclipse/IntelliJ settings files in the repo are not linked from the homepage or any of the guides
- The styleguides follow different conventions (xsl vs. css for styling, different versioning schemes (last changed vs revision number))
- cpplint.py would benefit from being hosted independently on github or similar 
(I forked to my personal github repo at https://github.com/tkruse/cpplint)
- Maintenance of the guides would be more convenient if the HTML/XML sources were written in a markdown dialect or reStructured text instead of xml. Possibly presenting via readthedocs would yield beautiful yet cheap results
- The Common LISP Guide contains General guidelines regarding any project, not just LISP
- code examples have no syntax highlighting
- all language specific guides would do well to include further configuration for the various existing static checker tools, such as checkstyle for Java or pep8 for python etc. This would work better if there were more of a folder structure.

In other words, the styleguide itself as an open-source project does not set a great example for how google should present open-source projects, IMO.

Compare the effort that goes into this project with the effort that went into material design guide: 
http://www.google.com/design/spec/material-design/introduction.html

Making the styleguide itself a model open-source project would easily be feasible as Summer of Code project or similar.

Original issue reported on code.google.com by `tibokr...@googlemail.com` on 6 Jan 2015 at 5:59

vapier:
there's a lot to unpack here.  in the future, please use sep issues for sep requests.  it's a pita to track multiple things in one report, and basically guarantees it stays open forever, and it's hard to have discussions.  as such, i'll close this out, and if you feel specific things still need addressing, please open a new report for each one.

> The project contains no README, LICENSE, CONTRIBUTING or CHANGELOG file (the project homepage contains that information instead, of course, but that's not how to do open-source, forks should have this information)

this is largely fixed with the GH migration, or at least as much as we plan on doing.

> The project homepage does not link to this issue tracker

it does now

> The homepage does not mention a maintainer, it is not clear at all who will respond to issues, and who may decide about style changes.

[...]

PR #82, opened 2015-10-26
replace xml/html files with asciidoc files for rendering on github

tkruse:
Since moving the styleguides from googlecode to github, the one written in xml have become unreadable. Those written in html have always been very bad to edit and fix.

This PR applies the transformation suggested by me in #36 (1st comment).


tkruse:
on the other hand, asciidoc does not serve as input to pandoc, not has github whitelisted the jekyll asciidoc plugin, so markdown is probably a better choice.

DJViking:
What happened to this change? Would not the use of Asciidoc instead of HTML be much better for future maintenance. You could build both HTML and PDF from Asciidoc.
We use the Google Java Code Standard, with some minor adjustment (indentation for instance). It would be much easier for use to copy the standard if it where written in Asciidoc.
We could however use Pandoc to produce Asciidoc from the HTML, but it does not produce a 100% correct output.

PR #87, opened 2015-11-09
Link to C++ Style Guide broken

ericcurtin:
Replaced .xml with .html


vapier:
thanks for the heads up.  looks like this is fixed in the latest version now.

Issue #79, opened 2015-10-14
check header guard variable for more extensions then "h"


dirk-thomas:
Currently the header guard variables are only checked for files ending with `.h` (https://github.com/google/styleguide/blob/2bfe5f4aa813e9294b3c3719d9648da90d84ac20/cpplint/cpplint.py#L4434).

When additional extensions are passed via the command line argument (e.g. `--extensions=c,cc,cpp,cxx,h,hh,hpp,hxx`) (some of) those should be checked too. This could either be done by passing the header extensions separately or consider all files which extension starts with `h`?

I am happy to provide a PR for this if there is a consensus how it should be addressed. I would suggest changing the referenced line to:

```
if file_extension.startswith('h'):
```

dirk-thomas:
A similar check is being used in five locations. They would all benefit from checking with `startswith()` instead.

LukeCz:
Added configurable option for any extension type in pull request #183.

dirk-thomas:
Thank you for creating the new pull request with the new header option. In order to cover the use cases described here and the other PR (#88) it would be good if your PR would also allow to specify an option to override the extension of source files.

LukeCz:
I fully agree, .cc should not be hardcoded either. I will add this option once/if this PR is merged, it should be easy to follow the same pattern :)
I do not want to code it immediately in case there are concerns regarding my approach and I will have to change.

tkruse:
@LukeCz: In case it helps, in our fork of cpplint similar work has been done to consistently handle extensions: https://github.com/theandrewdavis/cpplint

We would love to have our work included upstream, but so far nobody showed interest in merging open PRs.

LukeCz:
@tkruse: Ah, it seems we had a similar idea:) I would love for either one to be accepted to the official version so I do not have to merge the changes every time I download new version.

LukeCz:
@dirk-thomas I think .cc is already addressed in the code - _IsSourceExtension() function - and my recent commit addresses .h.
It seems to me this case can be closed successfully.

dirk-thomas:
@LukeCz I commented on the corresponding PR because I don't think the current code covers "enough" extensions: see https://github.com/google/styleguide/pull/88#issuecomment-250628631

vapier:
public cpplint development in this repository has been halted (see https://github.com/google/styleguide/pull/837). please migrate to https://github.com/cpplint/cpplint now.

PR #88, opened 2015-11-14
instead of checking for an exact extension (.h / .cc) only check that it starts with .h / .c

dirk-thomas:
This addresses #79.

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -4431,7 +4431,7 @@ def CheckStyle(filename, clean_lines, linenum, file_extension, nesting_state,
 
   # Check if the line is a header guard.
   is_header_guard = False
-  if file_extension == 'h':
+  if file_extension.startswith('h'):
     cppvar = GetHeaderGuardCPPVariable(filename)
     if (line.startswith('#ifndef %s' % cppvar) or
         line.startswith('#define %s' % cppvar) or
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Ping :wink:

LukeCz:
I think .cc is already addressed in the code - _IsSourceExtension() function - and my recent commit addresses .h.
It seems to me this case can be closed.

dirk-thomas:
There are further extensions beyond `c` and `cc` like `cpp` and `cxx`. So I don't think a static list is sufficient for these cases.

LukeCz:
@dirk-thomas I was checking the code and it seems to me that new code compares with the --extensions list instead of ='.cc' so once you pass cxx, cpp and any other that should work.

dirk-thomas:
I am sorry. I didn't know about the recently added `--headers` option. With `--headers` and `--extensions` all use cases of this PR are covered. Closing...

dirk-thomas:
@LukeCz Since you seem to be actively maintaining `cpplint` I want to ask if you would accept PRs for any of the patches we still maintain on-top of upstream?
- https://github.com/ament/ament_lint/commit/49d787a618088b3fd314619d3ba8aa91b7936f59
- https://github.com/ament/ament_lint/commit/b8ab0e6e5ba19dc99ab95b5db3d52a449ee7a2cc
- https://github.com/ament/ament_lint/commit/7f2af60114d1aad63445eb52032fca0db380e042
- https://github.com/ament/ament_lint/commit/1700c7ebc96bf8288f62843e8f9a2a8134c4a71c

For some it would be necessary to add command line options. I just don't want to spend the time to create PRs if they won't be considered anyway.

LukeCz:
@dirk-thomas I'm sorry for the confusion but I do not have wrights to accept. I just posted my PRs for the features that were missing for me. @eglaysher has been active recently merging changes so maybe you could ping him about your changes.

PR #92, opened 2015-12-04
Using namespace literals

mattyclarkson:
In C++14 new user defined literals are, by convention, contained in a namespace `literals`:

```
#include <iostream>
#include <chrono>
int main()
{
    using namespace std::chrono_literals;
    auto day = 24h;
    auto halfhour = 0.5h;
}
```

We should allow these types of `using` declarations.


tkruse:
See https://github.com/tkruse/cpplint/tree/using-namespace-literals

geoffromer:
The C++ style guide's ban on `using namespace` has no exception for literal namespaces, so cpplint shouldn't either.

PR #96, opened 2015-12-22
Modernize IntelliJ style file

dflemstr:
This removes tags from the style file that are no longer being read by
IntelliJ, and adds them back in the new format.  Without this change,
it's not even possible to use the style file in IntelliJ 15 or later,
since most of the old tags are not being read by the IDE.


Endron:
+1

rdifalco:
+1

PR #99, opened 2016-01-06
Update to current.

tituswinters:
Most significant updates:
- Casting (for arithmetic types, maybe use {}s)
- Namespaces
- Aliases

(I'll ask the other arbiters to double check as well if you like)


geoffromer:
LGTM

PR #100, opened 2016-01-07
Make cpplint less hardcoded to '.cc' (e.g. support '.cpp')

adsk-mage:
This makes it possible to run cpplint on files called foo.cpp/foo.h (not only foo.cc/foo.h).

Previously .cpp files were partially supported, but build/include_what_you_use in particular
would give false positives since a .cpp file would not be treated as a .cc file.

In addition, a couple of Windows related bugs were fixed (handling of '\' vs '/' in paths).

@tituswinters , @eglaysher - is a GitHub pull request the right method for a cpplint patch?

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -1674,7 +1674,7 @@ def GetHeaderGuardCPPVariable(filename):
   fileinfo = FileInfo(filename)
   file_path_from_root = fileinfo.RepositoryName()
   if _root:
-    file_path_from_root = re.sub('^' + _root + os.sep, '', file_path_from_root)
+    file_path_from_root = re.sub('^' + _root + '/', '', file_path_from_root)
   return re.sub(r'[^a-zA-Z0-9]', '_', file_path_from_root).upper() + '_'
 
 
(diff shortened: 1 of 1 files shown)


tituswinters:
Lemme poke at it a little bit.

eglaysher:
Sending a pull request is fine. Also, have you signed the CLA? (https://cla.developers.google.com/clas)

adsk-mage:
@eglaysher - I"ll have to check our CLA status.

cpplint/cpplint.py
```diff
@@ -1670,7 +1674,7 @@ def GetHeaderGuardCPPVariable(filename):
   fileinfo = FileInfo(filename)
   file_path_from_root = fileinfo.RepositoryName()
   if _root:
-    file_path_from_root = re.sub('^' + _root + os.sep, '', file_path_from_root)
+    file_path_from_root = re.sub('^' + _root + '/', '', file_path_from_root)
```
ArekSredzki, line 1677: Why did you remove the os specific separator?

tkruse, line 1677: I think this has been merged with a fix for this. PR should be close, I believe (3b0ea894c11140c1, also from 01e47236c84 "444 - Reduced dependency on hardcoded .cc extension.") Also see the fork at https://github.com/theandrewdavis/cpplint for various other fixes around extensions. As an example, the cpplint.py code here still checks for .h extension only in CheckForHeaderGuard, not for hpp and other header extensions used in other places in cpplint.

vapier:

PR #108, opened 2016-03-07
Don't replace - with _ in header names prior to sorting.

jeisinger:
The styleguide doesn't have this rule and it conflicts with what
clang-format does.

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -667,7 +667,6 @@ class _IncludeState(object):
   def CanonicalizeAlphabeticalOrder(self, header_path):
     """Returns a path canonicalized for alphabetical comparison.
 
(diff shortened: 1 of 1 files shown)


tituswinters:
IIUC, the current behavior is already compatible with clang-format, in that it
accepts either sort order:
(legacy)
foo.h
foo-inl.h

(new)
foo-inl.h
foo.h

jeisinger:
which commit of cpplint.py accepts both behaviors? In V8 we use ToT clang-format and the version of cpplint.py checked into chromium, and they disagree on the include order...

After talking with Daniel Jasper, he pointed out that replacing - with _ was the culprit (cf https://bugs.chromium.org/p/chromium/issues/detail?id=591952)

tkruse:

vapier:

PR #109, opened 2016-03-07
CppLint: Make output appear in the Visual Studio Error List

SteveVanOpstal:
Visual Studio 2003 - 2015 keep following output format:

```
{filename (line# [, column#]) | toolname} : [any text] {error | warning} code####: localizable string [any text]
```

[source](https://msdn.microsoft.com/en-us/library/yxkt8b26%28v=vs.140%29.aspx)

In CppLint the `{error | warning}` part is missing causing the output to only appear in the 'Output' screen and not in the 'Error List'. This makes the warnings more difficult to detect. This pull request simply adds 'warning:' to the output string.


vapier:

PR #111, opened 2016-03-08
Rephrase the last paragraph in Local Variables section of cppguide.html to make it more clear

bearlin:
Rephrase the last paragraph in Local Variables section of cppguide.html to make it more clear

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -820,7 +820,7 @@ for (int i = 0; i &lt; 1000000; ++i) {
 </pre>
 
 <p>It may be more efficient to declare such a variable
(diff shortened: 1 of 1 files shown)


mzeren-vmw:
FWIW my reading of the guide is that it consciously keeps language informal. This change while clarifying also makes the language more formal. I might have simplified further:

"It may be more efficient to declare such variables outside the loop, as in:"

bearlin:
Good idea, this make it more clear to me, I will update it to "It may be more efficient to declare such variables outside the loop, as in:"

PR #115, opened 2016-03-24
Sync Java IDE styles with internal configuration

ejona86:
I don't know how well this new version interacts with newer IntelliJ and I also know it has some deficiencies like missing the "io" TLD for import sorting. But syncing from our internal version before trying to solve those issues should help in maintaining it.

### DIFF
--- a/intellij-java-google-style.xml
+++ b/intellij-java-google-style.xml
@@ -7,258 +7,11 @@
       <option name="TAB_SIZE" value="8" />
       <option name="USE_TAB_CHARACTER" value="false" />
       <option name="SMART_TABS" value="false" />
-      <option name="LABEL_INDENT_SIZE" value="0" />
-      <option name="LABEL_INDENT_ABSOLUTE" value="false" />
-      <option name="USE_RELATIVE_INDENTS" value="false" />
     </value>
   </option>
-  <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
-  <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
-  <option name="IMPORT_LAYOUT_TABLE">
-    <value>
-      <package name="com.google" withSubpackages="true" static="false" />
-      <emptyLine />
-      <package name="android" withSubpackages="true" static="false" />
-      <emptyLine />
-      <package name="antenna" withSubpackages="true" static="false" />
-      <emptyLine />
-      <package name="antlr" withSubpackages="true" static="false" />
-      <emptyLine />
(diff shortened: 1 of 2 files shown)


lukaszx0:
@ejona86 thanks! I've used this config but I had to adjust tab settings. This is what got saved in my xml after changing that manually in GUI:

```
  <codeStyleSettings language="JAVA">
    <indentOptions>
      <option name="INDENT_SIZE" value="2" />
      <option name="CONTINUATION_INDENT_SIZE" value="4" />
      <option name="TAB_SIZE" value="2" />
    </indentOptions>
  </codeStyleSettings>
```

(at the very bottom, before `</code_scheme>`)

ejona86:
@nmittler, would you mind reviewing? This is a copy of the internal IDE configurations.

intellij-java-google-style.xml
```diff
@@ -267,6 +20,132 @@
   <option name="ALIGN_MULTILINE_EXTENDS_LIST" value="true" />
   <option name="ALIGN_MULTILINE_PARENTHESIZED_EXPRESSION" value="true" />
   <option name="ALIGN_MULTILINE_ARRAY_INITIALIZER_EXPRESSION" value="true" />
+  <option name="KEEP_BLANK_LINES_IN_CODE" value="1" />
+  <option name="BLANK_LINES_AFTER_CLASS_HEADER" value="1" />
+  <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
+  <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="99" />
+  <option name="IMPORT_LAYOUT_TABLE">
```
nmittler, line 27: Any reason to move the IMPORT_LAYOUT_TABLE? It's hard to diff ... were there any changes?

intellij-java-google-style.xml
```diff
@@ -286,191 +166,9 @@
   <option name="DOWHILE_BRACE_FORCE" value="3" />
   <option name="WHILE_BRACE_FORCE" value="3" />
   <option name="FOR_BRACE_FORCE" value="3" />
-  <ADDITIONAL_INDENT_OPTIONS fileType="css">
```
nmittler, line 289: Do we no longer need the additional indent options?

punkeel:
Is this the final version, or not just yet?

ejona86:
@PunKeel, that was the intention. I was trying to sync with the google-internal IntelliJ style, but I think I ended up syncing from the wrong file internally. I needed to try testing out the other file I found and haven't had time.

The idea was that after I do a sync, I can then make changes internally and externally at the same time. (For things like #120)

PR #120, opened 2016-04-08
Adding io.* to IMPORT_LAYOUT_TABLE in IntelliJ style

lucastsa:


ejona86:
Adding `io.*` will be the first thing I do after #115. I need it as well for my project :)

lucastsa:
I signed it!

punkeel:
Just wondering. What's the criteria for TLD's to be introduced? Why isn `fr` absent, for instance?

ejona86:
The criteria was approximately "the ones we found in our code and OSS we were using." Once we're able to add `io`, it would be easy to add `fr`.

PR #125, opened 2016-04-25
Fixed a typo in pyguide.html

carolynfischer:
Renamed iff to if.


carolynfischer:
I signed it!

carolynfischer:
Signed it with a different account as well.

Stephan202:
@carolynfischer, I suspect the original author really did mean to say _iff_, meaning "[if and only if](https://en.wikipedia.org/wiki/If_and_only_if)".

carolynfischer:
Fair enough. :)

PR #130, opened 2016-05-06
Update cpplint.py to r456

alexv1n:
456 - Tweak lint to sometimes allow { on line following array initialization.
455 - Recognize more types.
454 - Try a bit harder to detect templated types.
453 - Changed error message wording on build/storage_class to be less ambiguous.
452 - Recognize more types to silence false positives for brace warnings.
451 - <skipped>
450 - <skipped>
449 - Ignore whitespace/tab for Linux Kernel files.
448 - Remove some more false positive lint warnings for int64{n}.
447 - Just one warning message for line length is enough, don't need two.
446 - Allow braced conversions to int16, uint64, etc. per the style guide.
445 - Fixed handling of backslash escapes for checking whether a "//" is quoted.
444 - Reduced dependency on hardcoded .cc extension.
443 - <skipped>
442 - Fix matching of macro names in CheckTrailingSemicolon to include digits.
441 - Deleted checks:
      - All checks for RValue references.
      - Spacing check around boolean &&, because those look like RValue
        references.
      A huge amount of code and effort were dedicated to tell RValue
      references apart from boolean expressions, and to differentiate
      allowed versus banned RValue references. But we still get regular
      false positives from this one check. Rather than making the check
      more complex than what it already is, it seemed safer to just delete
      it altogether for now, and have a different process for catching
      RValue references.
440 - Add rule to cpplint to throw error on empty if statement bodies without
      else clauses. If statement bodies with comments are not considered empty.
439 - Allow spaces before closing brace of namespace block so that namespaces
      inside of macro definitions (where the entire macro definition is
      indented) are not treated as errors.
438 - cpplint: fix a false positive on `new const int(x)`.
437 - Only check for function length when current line is inside a function.
436 - cpplint: Catch static `std::string` instances as well as those written as
      `string`. Previously users would sometimes work around the lint warning by
      changing their code to even worse code by adding the "std::" prefix.
435 - cpplint: Be a little smarter about warning on indentation.
434 - cpplint: Remove false positives on some functions returning string const&.
433 - cpplint: improve diagnostics of global/static string objects.
432 - Allow non-const reference parameters for iostream based types.
431 - Better handling of raw strings inside comments.
430 - <noop>
429 - <skipped>
428 - <skipped>
427 - Add support to CHECK_NOTNULL when checking if a member variable is
      initialized with itself.
426 - Allow suppressing specific warnings in C headers.
425 - Allow spaces before parens for inline assembly.
424 - Remove lint checks for {EXPECT,ASSERT}_.*_M.
423 - Allow comment lines of more than 80 characters if they contain a single
      "word" (without any spaces).
422 - cpplint: Warn on inclusion of C++14 headers.
421 - cpplint: recognize <shared_mutex> as a standard library header.
420 - Add <scoped_allocator> to cpplint's list of C++ standard headers.
419 - Add cpplint check for tr1/ headers. We removed the nanny guards for these
418 - <skipped>
417 - Update the styleguide and cpplint to allow unnamed parameters
416 - Remove the rule explicitly banning non-default move operations.
415 - Remove the check for explicit multi arg constructors
414 - Include clarity of lambda default captures in "pros" section, mention
      of capturing `this` and lambdas which escape the current scope in "cons".
      Soften the ban on default captures to a preference, with admonition against
      using them for capturing `this` or when they will escape the current scope.
      This is a fairly straightforward change with a brief inclusion
      of 2 problematic cases for implicit capture. As this is the style
      guide, I'm not sure how much more detail is appropriate.
413 - Fixed a bug by making regex in CleanseRawStrings match multiple raw strings
      on a single line in left-to-right order.
412 - Fixed false positive for classes derived using decltype()
411 - <skipped>
410 - Recognize '1LL<<20' as numeric and don't flag it for spacing.

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -177,6 +177,8 @@ Syntax: cpplint.py [--verbose=#] [--output=vs7] [--filter=-x,+y,...]
 _ERROR_CATEGORIES = [
     'build/class',
     'build/c++11',
+    'build/c++14',
+    'build/c++tr1',
     'build/deprecated',
     'build/endif_comment',
     'build/explicit_make_pair',
@@ -196,7 +198,6 @@ _ERROR_CATEGORIES = [
     'readability/check',
     'readability/constructors',
     'readability/fn_size',
-    'readability/function',
     'readability/inheritance',
     'readability/multiline_comment',
     'readability/multiline_string',
@@ -227,6 +228,7 @@ _ERROR_CATEGORIES = [
     'whitespace/comma',
     'whitespace/comments',
     'whitespace/empty_conditional_body',
+    'whitespace/empty_if_body',
     'whitespace/empty_loop_body',
     'whitespace/end_of_line',
     'whitespace/ending_newline',
@@ -245,6 +247,7 @@ _ERROR_CATEGORIES = [
 # compatibility they may still appear in NOLINT comments.
 _LEGACY_ERROR_CATEGORIES = [
     'readability/streams',
+    'readability/function',
     ]
 
 # The default state of the category filter. This is overridden by the --filter=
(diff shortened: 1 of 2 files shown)


alexv1n:
I signed it!

ahmdrsl417-art (approved):

PR #131, opened 2016-05-11
添加Google Java Style中文版

laibinzhi:


vapier:
sorry, but i don't think we're planning on hosting translations at this time

PR #134, opened 2016-05-16
Changed stream from stderr to stdout for successful processing

darkclouder:
Successful processing of files spams stderr, should be stdout because it's no error.
It also makes it easier to filter those by redirecting it to /dev/null


darkclouder:
I signed it!

tkruse:

vapier:

PR #135, opened 2016-05-19
Fix broken link

xyproto:
The link at the bottom of https://google.github.io/styleguide/shell.xml is
broken. When changing from `cppguide.xml` to `cppguide.html`, it works.

Also removed trailing whitespace.


eglaysher:
I do apologize for the delay in getting to this; previously, no one was currently watching this repository. Thank you for your contribution.

PR #137, opened 2016-05-20
Bug-fix: NOLINT, NOLINTNEXTLINE has no effect for lines like "};".

piotr-semenov:
Unfortunately, CppLint always prints out the error for "};" regardless of NOLINT, NOLINTNEXTLINE. Please find the minimal not-working example below:

```
1: #include <cstdlib>
2: #include <iostream>
3:
4: int main(int argc, char** argv) {
5:   for (std::size_t i = 0u; i != 100u; ++i) {
6:     std::cout << i << std::endl;
10:    // NOLINTNEXTLINE
11:    };  // NOLINT
12:
13:    return EXIT_SUCCESS;
14: }
```

CppLint unexpectedly prints out the error _You don't need a ; after a }  [readability/braces] [4]_ for line 11 from CPP above.

This commit fixes this bug.

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -3884,6 +3884,14 @@ def CheckTrailingSemicolon(filename, clean_lines, linenum, error):
       # outputting warnings for the matching closing brace, if there are
       # nested blocks with trailing semicolons, we will get the error
       # messages in reversed order.
+
+      # We need to check the line forward for NOLINT
+      raw_lines = clean_lines.raw_lines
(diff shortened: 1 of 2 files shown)


eglaysher:
(Sorry for the wait; I don't actually get notifications from this repository.)

Please add a unit test to testErrorSuppression() in cpplint_unittest.py for this case; since these test cases are just fragments of c++, you can probably reuse part of the fragment you put in the pull request.

piotr-semenov:
I've recently added the appropriate unit-test to cpplint_unittest.py. Please, review.

eglaysher:
Thank you for your patch.

PR #139, opened 2016-05-30
Fix regex escape issue when using --root on Windows

sergeyvfx:
The code was using directory separator in the regex itself,
which means on Windows it was leaving unescaped backslash.

Added explicit escape for such cases.


eglaysher:
I do apologize for the delay in getting to this; previously, no one was currently watching this repository. Thank you for your contribution.

PR #144, opened 2016-06-10
pyguide -- Fix broken URL

rayglover-ibm:


rayglover-ibm:
It's unlikely that CLA will get signed in the next few months. :( Since URLs aren't protected by copyright, I don't see any problem with someone taking this patch and committing it themselves.

Foxytine31:
I signed it

Foxytine31 (changes requested):
I signed it!

gpshead:
Thanks.  fyi - I will be pushing a big update to the python style guide soon as we unfortunately haven't kept up on updates to in here in our external repo for several years and need to fix that. This link will be cleaned up as part of that. :)

gpshead:
fix pushed.

PR #149, opened 2016-06-24
Update shebang line

alanyee:
This shebang line is more portable, as the user may not have python installed in /usr/local/bin/python. Also, according to PEP 486, the line makes the Python Launcher aware of virtual environments.


alanyee:
Not my first contribution to a Google open source project, but I signed it anyway!

eglaysher:
Thank you for your contribution.

PR #154, opened 2016-07-07
Add --quiet command line argument

sergeyvfx:
Quiet makes output quiet unless errors occurs.

Mainly used by automation tools when parsing huge amount of files.
In those cases actual error might get lost in the pile of other stats
prints.

This argument is also handy for build system integration, so it's
possible to add automated lint target to a project and invoke it
via build system and have no pollution of terminals or IDE.

For example when hooking cmakelint to a Ninja project without this
argument one will have screen filled with "Total Errors: 0" messages,
which defeats all the idea of clean Ninja output.


vapier:

PR #155, opened 2016-07-07
Consider foreach a known loop structure

sergeyvfx:
Without this using C++11's foreach triggers a false positive
about extra space after foreach since linter considers this
a function call.

This is a bit controversial, and perhaps should be moved to
a filter somehow since in theory it might conflict with
pre-C++11 code which might have foreach declares as a regular
function.

Hope to find some solution which will satisfy everyone :)


tituswinters:
There is no "foreach" in C++11.  It's still spelled "for".

for (const auto& item : container)

PR #169, opened 2016-08-11
Fix the `cpplint.py` `build/endif_comment` check.

arames:
- The check needs to be run before we remove comments, otherwise valid
  lines will be found as invalid.
- A single character different from `/` after the spaces is enough to
  indicate an error.
- Also catch errors when only one `/` is present.
  (For example `#endif / MY_FILE_H`.)

Tested with:

```
// Correct

#endif// FOO_1
#endif // FOO_2
#endif  // FOO_3
#endif  //
#endif  // blargh
#endif  // one tab
#endif      // two tabs

// Incorrect
// Note that there are spaces and tabs below.
#endif
#endif 
#endif          
#endif  
#endif                          
#endif x
#endif xxx
#endif / xxx
```

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -2683,6 +2683,11 @@ def CheckForNonStandardConstructs(filename, clean_lines, linenum,
            filename, line number, error level, and message
   """
 
+  line = clean_lines.lines_without_raw_strings[linenum]
+  if Match(r'\s*#\s*endif\s*([^/\s]|/[^/]|$)', line):
+    error(filename, linenum, 'build/endif_comment', 5,
+          'Uncommented text after #endif is non-standard.  Use a comment.')
+
   # Remove comments from the line, but leave in strings for now.
   line = clean_lines.lines[linenum]
 
@@ -2712,10 +2717,6 @@ def CheckForNonStandardConstructs(filename, clean_lines, linenum,
     error(filename, linenum, 'build/storage_class', 5,
           'Storage-class specifier (static, extern, typedef, etc) should be '
           'at the beginning of the declaration.')
-
-  if Match(r'\s*#\s*endif\s*[^/\s]+', line):
-    error(filename, linenum, 'build/endif_comment', 5,
-          'Uncommented text after #endif is non-standard.  Use a comment.')
 
   if Match(r'\s*class\s+(\w+\s*::\s*)+\w+\s*;', line):
     error(filename, linenum, 'build/forward_decl', 5,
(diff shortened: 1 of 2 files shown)


arames:
I should be on the CLA as alexandre.rames@gmail.com.

arames:
Updated the CLA.

eglaysher (changes requested):
(Thank you for pinging me. Nobody actually is notified on the pull request queue.)

The main change looks good, but could you please put the test cases you have in the conversation into cpplint_unittest.py? You should be able to just add them to testBuildEndComment().

arames:
Thanks for the review. I have updated the unit tests, fixing what was necessary.

eglaysher (approved):

eglaysher:
Thanks!

dirk-thomas:
This patch makes comments behind `#endif` mandatory. Before they were optional. As far as I read the styleguide it doesn't say that every `#endif` must have a comment, e.g. this code snippet should be perfectly valid https://google.github.io/styleguide/cppguide.html#Preprocessor_Directives

dirk-thomas:
This patch makes comments behind `#endif` mandatory. Before they were optional. As far as I read the styleguide it doesn't say that every `#endif` must have a comment, e.g. this code snippet should be perfectly valid https://google.github.io/styleguide/cppguide.html#Preprocessor_Directives

**Update:** I just found that this has been reported in #186.

eglaysher:
Done.

arames:
"Before they were optional." That is only because the code was plainly broken and _never_ checking the `#endif` for comments.

I see multiple options:
- Make this a warning instead of an error. (Not sure whether cpplint has warnings.)
- Do not fix this. Then we should _remove_ this check and the related code, as it is currently not exercised.
- Reapply the fix. Then users should decide to not enable the associated check. That is what we did in our project, as we also felt that commenting after every `#endif` was too much.

arames:
Extending on my last bullet point. Currently having the check enabled does nothing. My patch was restoring the functionality that the check was designed to do. After the fix users need to decided whether they want this behaviour or not (which as far as I saw was the intended behaviour) by enabling or disabling the `build/endif_comment` check in their config.

LukeCz:
I agree with @arames. This behavior should be controlled by build/endif_comment check. This way anyone can decide if this check is needed or not. The same with multiple other checks that may be controversial for some people. I think the change should've not be reverted.

LukeCz:
@eglaysher, any thoughts on this issue?

arames:
I uploaded a revert of the revert at
https://github.com/google/styleguide/pull/193

PTAL

PR #172, opened 2016-08-19
Java indent configuration by fileType use JAVA key instead of java

nhenneaux:
The filetype java is not working anymore with Intellij 2016.2.


nhenneaux:
I signed it!

PR #175, opened 2016-09-03
Extended CStyleCast regular expressions

calcmogul:
The current static_cast regular expression now also checks for "unsigned", "unsigned char", "unsigned int", and the types in stdint.h (e.g., uint8_t). The reinterpret_cast regular expression now also checks for const qualified pointers.

The unit test coverage for both regular expressions was also improved.


tkruse:

PR #180, opened 2016-09-19
Update htmlcssguide.xml

Kledal:
Add closing tag to p


Kledal:
I signed it!

hugovk:
@Kledal The same style guide says to omit optional tags like `</p>`

https://google.github.io/styleguide/htmlcssguide.xml?showone=Optional_Tags#Optional_Tags

Kledal:
The documentation for when its allowed to omit p tag can be found [here](https://html.spec.whatwg.org/multipage/semantics.html#the-p-element).

I'm closing this PR as the example **is** correct and the p tag is allowed to be omitted.

Issue #181, opened 2016-09-24
Cpplint is unnecessarily polluting stderr


LukeCz:
Hi,

Cpplint is too verbose when integrating with build tools like make. Messages "Total errors found: " and "Done processing " are displayed even when there are no errors, polluting the screen. Those messages should go to stdout and not stderr so they can be redirected to /dev/null.

PR #182, opened 2016-09-24
Cpplint is unnecessarily polluting stderr

LukeCz:
This solves #181.
Allows to separate errors and normal output into separate streams.
Useful when standard output is to be redirected to /dev/null leaving
only errors.

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -2683,6 +2683,11 @@ def CheckForNonStandardConstructs(filename, clean_lines, linenum,
            filename, line number, error level, and message
   """
 
+  line = clean_lines.lines_without_raw_strings[linenum]
+  if Match(r'\s*#\s*endif\s*([^/\s]|/[^/]|$)', line):
+    error(filename, linenum, 'build/endif_comment', 5,
+          'Uncommented text after #endif is non-standard.  Use a comment.')
+
   # Remove comments from the line, but leave in strings for now.
   line = clean_lines.lines[linenum]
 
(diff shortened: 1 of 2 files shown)


LukeCz:
I signed it!

eglaysher:
I'm not convinced that this is worth having. I don't see the majority of people using this option.

LukeCz:
@eglaysher: So maybe let me ask from a different angle - what is the reason all is put to stderr and not to appropriate streams?
Let's take an example, in brackets are my suggestions on what should be on each stream. If we can agree on the default I can make a simple PR to just change the default instead of configurable option. Basically when you're scanning hundreds of files you don't want to have your screen clogged with noise.

`h.hpp:0:  No copyright message found.  You should have a line: "Copyright [year] <Copyright Owner>"  [legal/copyright] [5]` (this should be on **error** stream)
`Done processing h.hpp` (this should be on **standard** stream, it's not an error)
`Total errors found: 1` (this should be on error stream **only if** number of errors > 0)

eglaysher:
I'm mostly against having this as an option. The overwhelming majority of people will never reach for options that don't immediately solve a problem that they are having right now; if you're code base is using 'hpp' as an extension for headers, you are not going to not use cpplint unless you have some sort of option; things not working will redirect the user to the --help. Things being slightly annoying won't.

If this is the correct thing to do, it should be done by default.

That said, I'm  worried about this breaking tools which already exist and consume cpplint's output. I checked chromium's git-cl, but that appears to just use cpplint as a library and wouldn't be broken by a defaults change as above.

LukeCz:
Hi @eglaysher, I changed the defaults to what I think makes most sense - errors separated from end statements. This way "Total" and "Done" can be easily filtered out if I'm interested only in errors. Let me know what you think :)

eglaysher (approved):
I'm still a little worried about breaking existing cpplint users by changing this, but I suppose this is small compared to some of the other changes we've pushed.

eglaysher:
Actually, one final thing. Could you please squash this into a single commit? There's a lot of history gunk here that probably shouldn't live in the repository Forever.

LukeCz:
@eglaysher I was trying to squash those commits but after long time only managed to delete too much. I started new PR #184 with the same code. I would appreciate if you could merge #184.

PR #184, opened 2016-09-29
Cpplint is unnecessarily polluting stderr

LukeCz:
This solves #181.
Allows to separate errors and normal output into separate streams.
Useful when standard output is to be redirected to /dev/null leaving
only errors.


eglaysher:
Thanks!

LukeCz:
Thanks @eglaysher

PR #192, opened 2016-10-06
Override indent spaces

leonidasIIV:
Added the Ability to override the default spaces per indent of 2 so that the end user is able to test against a different number of spaces per indent on the leading edge.


leonidasIIV:
I authored these commits.

Foxytine31 (commented):
Merge

tkruse:

PR #193, opened 2016-10-12
Revert "Revert "Fix the `cpplint.py` `build/endif_comment` check.""

arames:
This reverts commit ae11811f8191818b894707223c37bf27f7b9b729.

As described in pull request 169, users should chose to enable or
disable this check using `build/endif_comment`. As the code is, that
check is never exercised.

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -2710,6 +2710,11 @@ def CheckForNonStandardConstructs(filename, clean_lines, linenum,
            filename, line number, error level, and message
   """
 
+  line = clean_lines.lines_without_raw_strings[linenum]
+  if Match(r'\s*#\s*endif\s*([^/\s]|/[^/]|$)', line):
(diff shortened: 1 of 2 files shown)


kernc:
:-1: 

This check is within `class CheckForNonStandardConstructs`. Empty `#endif` lines _are_ standard.

The check should only prevent non-comment text after `#endif`, as the [comment](https://github.com/google/styleguide/blob/ae11811f8191818b894707223c37bf27f7b9b729/cpplint/cpplint.py#L2695) says. But comments are always allowed as they are [stripped sooner in the pipeline](https://stackoverflow.com/a/3461759/1090455).

This change will break many projects, leading most to disable this check, resulting in non-standard C/C++ code.

tkruse:
Ah, sorry, I missed the revert commit somehow. You are right, it is reverted on upstream, will release 1.4.3 in a moment with the revert.

vapier:

PR #196, opened 2016-11-03
Update Java and C++ style guides with various changes since last update

shicks:


shicks:
Linking to the internal style groups to make sure there's buy-in before we merge this.

cppguide.html
```diff
@@ -565,61 +569,23 @@ <h3 id="Namespaces">Namespaces</h3>
...
 <ul>
-  <li>
+  <li>Follow the rules on <a href="#Namespace_Names">Namespace Names</a>.
+  </li><li>Terminate namespaces with comments as shown in the given examples.
+  </li><li>
 
   <p>Namespaces wrap the entire source file after
   includes,  
-  <a href="https://gflags.github.io/gflags/">
+  <a href="http://google-gflags.googlecode.com/">
```
mzeren-vmw: I think we want the LHS github URL.

cppguide.html
```diff
@@ -1594,50 +1599,30 @@ <h3 id="Access_Control">Access Control</h3>
 be <code>protected</code> when using
 
 
-<a href="https://code.google.com/p/googletest/">Google
+<a href="http://code.google.com/p/googletest/">Google
```
mzeren-vmw: Do we really want to drop https? Double check other URLs in this edit. Also this particular URL should be https://github.com/google/googletest.

cppguide.html
shicks: This has been reverted. The internal version of this file had gotten out of sync.

cppguide.html
shicks: Same here - this should be fixed now.

mzeren-vmw: Looks good.

geoffromer:
LGTM (C++ only, didn't look at Java)

tkruse:
BTW, without wanting to spoil all the fun of manually editing HTML, maybe any of the google maintainers would be interested in converting all guidelines to markdown? As I sketched here a long time ago: #36

shicks:
Internally the new JS guide _is_ markdown. Unfortunately, or internal
markdown renderer is a flavor of hoedown, whereas github pages uses
kramdown, so we end up converting it to HTML on the way out anyway...

tkruse:
Interesting. But still how is html better than hoedown on github?

shicks:
Because we can actually produce it and have it render correctly. Our style
guides take significant advantage of one of the nonstandard markdown
features (classes on code blocks and headings) which is unfortunately
incompatible between the two engines. So we'd need to do some sort of
postprocessing on it in any case, which is more moving parts that don't
exist yet and would need maintenance. Additionally, our externalization
process only works on HTML, so we'd need to devise a way to scrub markdown
directly, which also doesn't exist yet.

It's a great feature request to work toward, and I'd personally love to see
it happen, but given that our style guides are all volunteer/20% efforts,
it's not likely to happen in the near term.

On Nov 5, 2016 9:58 PM, "Thibault Kruse" notifications@github.com wrote:

Interesting. But still how is html better than hoedown on github?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
https://github.com/google/styleguide/pull/196#issuecomment-258660720, or mute
the thread

tkruse:
So this "externalization process" cut's away pieces that must not be open-source for now? 

Or could Google publish the raw hoedown and externalization scripts on github as well as the HTML so that contributors may both contribute to the raw sources directly as well as work on the transformation?

Additionally, what about the guides provided as xml files (objc, shell, lisp, ...), are they internally in hoedown, or is it interesting for them to be converted to markdown?

Foxytine31 (approved):

shicks:
Internal Java people are okay with this.  Can somebody merge it?

PR #200, opened 2016-11-17
insert missing word

clashman:


clashman:
I signed it!

geoffromer:
There's no missing word here; it's phrased as intended. "Takes some getting used to" is a fairly common English idiom.

Issue #201, opened 2016-11-17
Remove default list params in `cpplint.py`


erip:
Anywhere you have something like 

    def foo(bar=[]):
        # ...

It should probably be converted to 

    def foo(bar=None):
        # ...

While there's no mutation happening _within_ the functions using these default params, an adversary could still mutate them easily. See below:

```
>>> def foo(mut=[]):
...    # No mutation! We're safe... right?
...    print(mut)
...
>>> foo()
[]
>>> import inspect
>>> inspect.getargspec(foo)
ArgSpec(args=['mut'], varargs=None, keywords=None, defaults=([],))
>>> inspect.getargspec(foo).defaults[0].append(5)
>>> foo()
[5]
```

More details [here](http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments).

vapier:
public cpplint development in this repository has been halted (see #837).  please migrate to https://github.com/cpplint/cpplint now.

PR #202, opened 2016-11-17
Remove default list kwarg in `cpplint.py`

mrecachinas:
As identified in #201, this change removes the default list params in several functions and prefers the paradigm:

    def foo(x=None):
        if x is None:
            x = []
        ...


mrecachinas:
I signed it!

vapier:

PR #209, opened 2016-12-22
Update the IntelliJ style to the latest version from Google.

plumpy:
Internally, we use the google-java-format plugin to format our Java
files. It's much more rigorous and ignores the contents of this XML
file. Using that plugin should be the preferred way to make sure your
code conforms to Google's style.


shicks:
Thanks for preparing this!

Issue #174, opened 2016-08-31
False positive: Add #include <algorithm> for sort


rakhimov:
Please see https://github.com/theandrewdavis/cpplint/issues/27

mortenfyhn:
I have another very similar false positive:

Using a minimal example file `test.h` with contents

```cpp
// Copyright 2019 Morten Fyhn Amundsen

#ifndef TEST_H_
#define TEST_H_

void copy(int) {}

#endif  // TEST_H_
```

I also get a false positive suggesting to include the algorithm header:

```
$ cpplint/cpplint.py test.h
test.h:6:  Add #include <algorithm> for copy  [build/include_what_you_use] [4]
Done processing test.h
Total errors found: 1

```

I'm on the current version of the `gh-pages` branch (9c6b371).

---

More information:

```cpp
void copy() {}
```

gives no errors (no missing algorithm header message). However,

```cpp
void copy(void) {}
```

causes cpplint to report the missing algorithm header error.

vapier:
public cpplint development in this repository has been halted (see https://github.com/google/styleguide/pull/837). please migrate to https://github.com/cpplint/cpplint now.

PR #216, opened 2017-01-10
Fix false positive alarms on missed <algortihm> or <utility> header.

sergregory:
The case of a non-STL realization of the one of the functions under test that
is called via class name/namespace resolution syntax (e.g. type type::min)
was not handled by the regular expression.

Fixes the #174
@rakhimov, please check if this fix solves your case.


sergregory:
@googlebot, I signed it!

tkruse:

vapier:

PR #228, opened 2017-02-10
Use lowercase example in import

drigz:
This was probably there to illustrate the case insensitivity, but since importing classes is not allowed, this example is misleading.


gpshead:
I've got a major update to pystyle to push out based on our much more up to date internal version (the guide in this repo has been neglected for a few years). I just double checked, it includes cleaning this example up. no need to merge this change as it'll soon be obsolete.

PR #230, opened 2017-02-22
Teach the explicit constructor check about constexpr.

danakj:
If a constructor is marked constexpr it evades the explicit constructor
check right now, since the check only knows about the inline keyword.
Teach it that constexpr can be used also.


eglaysher:
Review: this looks fine, but please add a unit test. There's already a test testExplicitSingleArgumentConstructors(); just add a case in there to show that constexpr works.

danakj:
Thanks, I've added a bunch of tests, please have a look.

eglaysher:
LG. Thank you for being thorough on the tests.

PR #238, opened 2017-03-14
fix tag

al0mie:


ghost (approved):

shicks:
Seems clearer to me with the tag.  Otherwise it sounds like "required" just means "mandatory", whereas even without the `goog.` prefix, if it's in code font it's much more obviously talking about a dependency that was actually `goog.require`d.

But maybe others interpret it differently?  If it would read more clearly I'd rather add the `goog.` explicitly.  But I worry that it just muddies it up.

ghost (approved):

ghost (approved):

ghost (approved):

ghost (approved):

tonyruscoe:
FWIW, I think it reads fine without “required” at all given the context is already `goog.require` statements:

> These aliases are the only acceptable way to refer to the dependency, whether in code or in type annotations: …

vapier:
this section was dropped entirely in commit dceb47fb3ea99ad7cc4308fa2c9ecb0d012639e1

PR #241, opened 2017-03-30
cppguide: fix syntax error in lambda example

AleksandrKonstantinov:
Make an example with a lambda compilable.


AleksandrKonstantinov:
I signed it

PR #244, opened 2017-05-03
Changed vs7 output format so that Visual Studio can parse the error

schoetbi:
Also Visual Studio can jump to the correct file/linenumber


andlrc:
Doesn't VS have something like vim's [`errorformat`](http://vimdoc.sourceforge.net/htmldoc/quickfix.html#errorformat)?

schoetbi:
@andlrc: This is possible but I do not know. I changed the format in the py script the same way as the compiler issues the errors. That way the errors show up as if they are compiler errors and a double click right jumps to the erroneous line.

eglaysher:
Thanks!

PR #254, opened 2017-06-20
Fix indentation check for access keywords

enlich:
No whitespace before the access keywords in class/struct declarations
should be acceptable. The behavior of the check correctly reflects the
comment.


enlich:
I signed it!

vapier:

PR #256, opened 2017-06-22
Add Dart style guide

kwalrath:
* Added Go while I was in there. (@rakyll or @adg might have feedback on the link I used)
* Rewrote things a bit to make the new links fit.
* Copy edited a bit because I couldn't help myself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ which encourages you to share these documents.
 See [https://creativecommons.org/licenses/by/3.0/][ccl] for more details.
 
 The following Google style guides live outside of this project:
-[Go Style][go] and [Effective Dart][dart].
+[Go Code Review Comments][go] and [Effective Dart][dart].
 
 <a rel="license" href="https://creativecommons.org/licenses/by/3.0/"><img alt="Creative Commons License" style="border-width:0" src="https://i.creativecommons.org/l/by/3.0/88x31.png" /></a>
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,3 +49,6 @@ for more details.
 [cpplint]: https://github.com/google/styleguide/tree/gh-pages/cpplint
 [emacs]: https://raw.githubusercontent.com/google/styleguide/gh-pages/google-c-style.el
 [xml]: https://google.github.io/styleguide/xmlstyle.html
+[dart]: https://www.dartlang.org/guides/language/effective-dart
+[go]: https://github.com/golang/go/wiki/Style
```
rakyll: This is https://golang.org/wiki/CodeReviewComments now.

kwalrath: Yeah, I saw that. It's just a weirdly named page for a style guide, and I saw that Style linked to it. :) I'll update the link, though.

rakyll: Yeah, it is a bit weirdly named :) You can use the golang.org shortcut for the link btw: https://golang.org/wiki/CodeReviewComments. If we ever move the wiki somewhere, the link would still work.

kwalrath: SGTM. I'll do that now. Thanks for the feedback!

markmentovai: It looks like this list of links is ordered according to the sequence that things appear in the prose, but you inserted [go] before [dart] up there, and after it here.

kwalrath:
Thanks!

PR #265, opened 2017-07-24
Remove presubmit check for DISALLOW_* macros

backes:
Before C++11, we were using a hack to disable copy constructors or copy
assignment by declaring the methods private and not implementing them.
This hack required the respective macros to be placed in the private:
declarations of a class.
The macros have switched to use the C++11 "= delete" syntax some time
ago in both v8 and chromium:
https://codereview.chromium.org/1123723003/
https://codereview.chromium.org/2017213002

Also the comments are now updated, since the macros do not need to be
in the private: declarations any more:
https://chromium-review.googlesource.com/c/577687
https://chromium-review.googlesource.com/c/578027

This change also removes the presubmit check that checked that the
macros are only used in the private declarations.

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -3072,36 +3072,6 @@ def CheckComment(line, filename, linenum, next_line_start, error):
               'Should have a space between // and comment')
 
 
-def CheckAccess(filename, clean_lines, linenum, nesting_state, error):
-  """Checks for improper use of DISALLOW* macros.
-
-  Args:
-    filename: The name of the current file.
-    clean_lines: A CleansedLines instance containing the file.
-    linenum: The number of the line to check.
-    nesting_state: A NestingState instance which maintains information about
-                   the current stack of nested blocks being parsed.
-    error: The function to call with any errors found.
-  """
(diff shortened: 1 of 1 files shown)


vith:
Looks like the unit tests need to to be updated to reflect this change.

```console
$ python2 cpplint_unittest.py 
.......................................................................................................F....................................................................
======================================================================
FAIL: testMisplacedDisallowMacros (__main__.CpplintTest)
Traceback (most recent call last):
  File "cpplint_unittest.py", line 1987, in testMisplacedDisallowMacros
    '  [readability/constructors] [3]')
  File "cpplint_unittest.py", line 220, in TestMultiLineLint
    self.assertEquals(expected_message, self.PerformMultiLineLint(code))
AssertionError: 'DISALLOW_COPY_AND_ASSIGN must be in the private: section  [readability/constructors] [3]' != ''

Ran 172 tests in 0.547s

FAILED (failures=1)
```

eglaysher:
I went and fixed up the tests in 6c8669edb20249cd.

backes:
Thanks for fixing the tests, Elliot!

PR #268, opened 2017-08-23
[objc] Make objcguide.xml link to objcguide.md 🔄

stephanemoore:
There are links out in the wild to the old XML style guide. An XML file that links to the new style guide location ensures that those links are not completely broken.

### DIFF
--- a/objcguide.xml
+++ b/objcguide.xml
@@ -3,6 +3,6 @@
 <GUIDE title="Google Objective-C Style Guide">
   <p>
     The style guide has moved to
-    <a href="objcguide.md">objcguide.md</a>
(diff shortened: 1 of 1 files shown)


markmentovai:
This will be served at https://google.github.io/styleguide/objcguide.xml, so the relative link will point to https://google.github.io/styleguide/objcguide.md, which probably isn’t what you want (it’s raw, not rendered). You probably want https://github.com/google/styleguide/blob/gh-pages/objcguide.md.

stephanemoore:
Good catch 😅 I updated the link to https://github.com/google/styleguide/blob/gh-pages/objcguide.md.

markmentovai (approved):

stephanemoore:
Thanks for the review!

PR #271, opened 2017-09-05
Update eclipse-java-google-style

donce:
Remove blank lines between import groups to use [code style] (https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing).

A new problem occurs - static imports are not separated from non-static imports, but that's a smaller issue IMHO.


donce:
I signed it!

PR #278, opened 2017-09-23
Ignore file header formatting on Java

kazuki43zoo:
I want to ignore file header formatting on Java, because I add file header using build tool like the [formatter-maven-plugin](https://github.com/revelc/formatter-maven-plugin). (= I want to keep original format)
I read the "3.1 License or copyright information, if present](https://google.github.io/styleguide/javaguide.html#s3.1-copyright-statement)" section of "Google Java Style Guide", it does not mention formatting of the file header.

What do you think?


kazuki43zoo:
I signed it!

Issue #273, opened 2017-09-06
Java style for Eclipse formats imports incorrectly


donce:
From https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing:

>If there are both static and non-static imports, a single blank line separates the two blocks.

Now, all imports are in a single block:

```
import static org.hamcrest.CoreMatchers.is;
import static org.junit.Assert.assertEquals;
import static org.junit.Assert.assertFalse;
import java.net.URI;
import java.util.Collections;
import org.eclipse.core.runtime.CoreException;
import org.junit.Test;
```

Empty line should be left between static and non-static imports, so it should be like this:
```
import static org.hamcrest.CoreMatchers.is;
import static org.junit.Assert.assertEquals;
import static org.junit.Assert.assertFalse;

import java.net.URI;
import java.util.Collections;
import org.eclipse.core.runtime.CoreException;
import org.junit.Test;
```

This formatting is done by this line in config:
```
<setting id="org.eclipse.jdt.core.formatter.blank_lines_between_import_groups" value="0"/>
```

The problem is that Eclipse doesn't seem to support custom spacing between static and non-static groups.

nicolai86:
Is there a workaround available besides running the formatter outside of eclipse?

donce:
Using IntelliJ? :D

mkolesnikov:
For those who still comes to this issue for solution, here is how to Organize Import should be set up:
* In Organize Import, keep only 2 groups, one for all static and one for everything else
* In Formatter, Blank lines#Between import groups set 1. (the preview window does not organize imports, that's fine)
![image](https://user-images.githubusercontent.com/754163/51789700-caa5f780-219c-11e9-8bf3-e07b6d72ac08.png)
![image](https://user-images.githubusercontent.com/754163/51789771-93841600-219d-11e9-9a35-84bd9780f9bd.png)
Works in Eclipse Photon

MarceloLeite2604:
Thanks @mkolesnikov for the workaround. It helped me out organize import groups on my project.
I also used your second workaround to add a blank line between import groups.
I think GoogleStyle formatting profile defined on [eclipse-java-google-style.xml](https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml) should have this value defined, so it wouldn't be necessary to change it after import. Is it possible to add it?

oliviercailloux:
[...]

PR #279, opened 2017-09-23
Allow break line between import groups

kazuki43zoo:
Google Java Style Guide says as folows in "[3.3 Import statements](http://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing)":

```
Imports are ordered as follows:

1. All static imports in a single block.
2. All non-static imports in a single block.

If there are both static and non-static imports, a single blank line separates the two blocks. There are no other blank lines between import statements.
```

But the current `eclipse-java-google-style.xml` not allow adding the break line between import groups.


recena (approved):

Marcono1234:
Partially fixes #273
(the main problem for #273 is that the formatter config apparently cannot specify the import groups, so the user would have to edit them manually, as shown in https://github.com/google/styleguide/issues/273#issuecomment-457843194)

See also #271

florianlacreuse:
Hi,

Thanks for the fix @kazuki43zoo.

Any updates about the merge of this pull request (it's been 7 years)?

PR #283, opened 2017-10-10
Converted cpplint to python 3.

lielfr:

PR #284, opened 2017-10-11
Cpplint: Specify Python version

mariosal:
When executed with Python 3, it returns with error code `1`.


mariosal:
I signed it!

mariosal:
Possible overlap with #283

PR #286, opened 2017-10-17
Include run_prettify.js script and update all pre tags to use it

ArnaudBienner:
IMHO it would be nice to have the code snipped being colorized to (hopefully) make the code examples easier to read.
Turns out there is a Google project doing that :) 
https://github.com/google/code-prettify
I've added this script + added prettyprint class to all &lt;pre&gt; tag so they get colorized.

Might be worth to do the same thing for &lt;code&gt; but IMHO it is less needed as those are smaller portion of code + not all &lt;code&gt; tags are cpp, so special care should be taken when specifying classes on those. So let's worry about this in a second step.

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -6,6 +6,7 @@
 <link rel="stylesheet" type="text/css" href="include/styleguide.css">
 <script language="javascript" src="include/styleguide.js"></script>
 <link rel="shortcut icon" type="image/x-icon" href="https://www.google.com/favicon.ico" />
+<script src="https://cdn.rawgit.com/google/code-prettify/master/loader/run_prettify.js"></script>
 </head>
 <body onload="initStyleGuide();">
 <div id="content">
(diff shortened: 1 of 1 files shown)


ArnaudBienner:
I signed it!

shicks:
@tituswinters This change would need to be mirrored internally or else it will just get clobbered on the next push.

FWIW, JS and Java already both serve this same script for colorizing.

ArnaudBienner:
@shicks @tituswinters so, do you think this could be something worth to merge? Or do you have any other comments about this?

tituswinters:
Sorry about that - I don't monitor this repo closely since most of the traffic isn't relevant to me.  

Honestly, to resolve that we're looking at moving the C++ guide out of here and into the Abseil.io repository so that changes can be noticed more easily - it'll get converted to markdown in the process.  So I don't feel strongly either way.

@manshreck ?

ArnaudBienner:
@shicks @tituswinters any news about this?
I still think Google C++ styleguide is a useful resource, and would benefit from having color colorizing since it will make it easier to read (and will be consistent with other programming languages).

ArnaudBienner:
ping @pwnall since you've been active on this area recently.
I still believe this PR could be valuable.
@pwnall would you be interested in this change? I'm fine updating the merge request if that could help (though from what I understood those pages are generated automatically before being published in github).

pwnall:
@manshreck owns the decision for this. I'm fine either way, like Titus.

PR #287, opened 2017-10-27
Update intellij-java-google-style.xml

Tirex:
Default intellij make simple line comment on first position like this:
|//   some code
Accoding google styleguide same ident level must be using for comments too (https://google.github.io/styleguide/javaguide.html#s4.2-block-indentation)
This option disable  commenting on first column, so comment will look like:
|    //some code


Tirex:
I signed it!

2017-10-27 15:01 GMT+06:00 googlebot <notifications@github.com>:

PR #289, opened 2017-11-06
cpplint: Fix --root for non-subdirectories

iam:
Using cpplint.py --root with directories at a more outer level
than the repository would not work.

For example given

  $> cpplint.py  --root=/a/b  /a/b/c/.git/filename.cpp

Trying to use --root=/a/b would get ignored, and it would
ask for a headerguard of FILENAME_H_ instead of C_FILENAME_H_
as expected.

Now --root will always use the "full name" and then strip off --root
prefix, giving us the desired effect.

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -1777,6 +1777,8 @@ def GetHeaderGuardCPPVariable(filename):
   fileinfo = FileInfo(filename)
   file_path_from_root = fileinfo.RepositoryName()
   if _root:
+    file_path_from_root = fileinfo.FullName()
+
     suffix = os.sep
     # On Windows using directory separator will leave us with
     # "bogus escape error" unless we properly escape regex.


iam:
@eglaysher 

Hi, do I need to add any tests for this? We forked our own cpplint for Android and I'm trying to unfork it. 

https://android.googlesource.com/platform/art/+/master/tools/cpplint.py
https://android.googlesource.com/platform/tools/repohooks/+/refs/heads/master/tools/cpplint.py

madeleine-empirical:
I tried this out, since I was having a similar problem. It solved my problem, though slightly clumsily, since it doesn't work if a root path contains `..`

I'm left calling it like:
```
cpplint.py --root "$(cd .. ; /bin/pwd)"
```

markmentovai (approved):

iam:
@madeleine-empirical That's a good idea, and I've incorporated this suggestion in my next revision.

iam:
I created a new pull request #291 since I can't seem to reopen this one.

PR #291, opened 2017-11-08
cpplint: --root now supports non-subdirectories

iam:
Using ```cpplint.py --root``` with directories at a more outer level
will now prepend the header guard with all the directories from the
root to the file.

For example given
```
  ls /a/b/c # /a/b/c/.git /a/b/c/filename.h
  cpplint.py --root=/a/b /a/b/c/filename.h   # C_FILENAME_H_
```
This will prepend "C" to the normal header guard resulting in ```C_FILENAME_H_```
```
  # no root behavior:
  cpplint.py /a/b/c/filename.h               # FILENAME_H_
```
Also supports relative paths:

```
  cd /a/b/c
  cpplint.py --root=.. filename.h            # C_FILENAME_H_
```

Note that the old usage is still supported:
```
  cd /a/b/c
  mkdir -p d/e/f
  touch /a/b/c/d/e/f/filename.h
  cpplint.py --root=d/e/f d/e/f/filename.h   # FILENAME_H_
```
which would "strip" the prefix rather than prepend an extra prefix.

(Invalid root prefixes are as before also ignored)

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -1754,6 +1754,30 @@ def GetIndentLevel(line):
   else:
     return 0
 
+def PathSplitToList(path):
+  """Returns the path split into a list by the separator.
+
+  Args:
+    path: An absolute or relative path (e.g. '/a/b/c/' or '../a')
+
+  Returns:
+    A list of path components (e.g. ['a', 'b', 'c]).
+  """
+  lst = []
+  while True:
+    (head, tail) = os.path.split(path)
+    if head == path: # absolute paths end
+      lst.append(head)
+      break
+    if tail == path: # relative paths end
+      lst.append(tail)
+      break
+
+    path = head
+    lst.append(tail)
+
+  lst.reverse()
+  return lst
 
 def GetHeaderGuardCPPVariable(filename):
   """Returns the CPP variable that should be used as a header guard.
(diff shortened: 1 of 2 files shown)


iam:
```
$> ./cpplint_unittest.py 
...........................................................................................................................................................................
Ran 171 tests in 0.749s

OK
```

cpplint/cpplint.py
```diff
@@ -1776,13 +1776,37 @@ def GetHeaderGuardCPPVariable(filename):
...
     suffix = os.sep
     # On Windows using directory separator will leave us with
     # "bogus escape error" unless we properly escape regex.
     if suffix == '\\':
       suffix += '\\'
-    file_path_from_root = re.sub('^' + _root + suffix, '', file_path_from_root)
+
+    # root behavior:
+    #   --root=subdir , lstrips subdir from the header guard
+    (maybe_path, subs_made) = re.subn('^' + _root + suffix, '', file_path_from_root)
```
markmentovai: Please don’t do this, it’s dangerous. There are alternatives to re that may work. If you’d like to stick with re, take a look at re.escape().

Doing this correctly should eliminate the need for the “+= '\\'” a couple of lines above.

Again on line 1802.

markmentovai (changes requested):
The use of unescaped regexes is dangerous. I know that the existing code did it, but since you’re changing it, you should take the opportunity to make things better.

iam:
Removed usages of unescaped regexes, thanks for pointing this out!

cpplint/cpplint.py
```diff
@@ -1776,13 +1800,37 @@ def GetHeaderGuardCPPVariable(filename):
...
+    def StripListPrefix(lst, prefix):
+      # f(['x', 'y'], ['w, z']) -> None  (not a valid prefix)
+      if lst[:len(prefix)] != prefix:
+        return None
+      # f(['a, 'b', 'c', 'd'], ['a', 'b']) -> ['c', 'd']
+      return lst[(len(prefix)):]
+
+    # root behavior:
+    #   --root=subdir , lstrips subdir from the header guard
+    maybe_path = StripListPrefix(PathSplitToList(file_path_from_root), PathSplitToList(_root))
```
markmentovai: Style: stay within 80 columns. (I think that’s still Google style for Python? It certainly is Chromium style.)

Line 1826, and test lines 4255, 4257, 4267 too.

markmentovai (changes requested):
In the change description, the summary line can be improved to be more active:

cpplint --root: support non-subdirectories

iam:
```
$> ./cpplint_unittest.py 
............................................................................................................................................................................
Ran 172 tests in 0.836s

OK
```

markmentovai (approved):

PR #292, opened 2017-11-09
CPPLINT.cfg root=setting is now relative to CFG file.

iam:
Also further improve the documentation and testing around ```--root```.

Previously the `root=setting` in the CFG file was treated identically
to passing ```--root=setting``` on the command line, which seems undesirable
since it depends on were CPPLINT.cfg was invoked from (for relative
paths).

Judging on settings such as ```'exclude_files'``` it seems within the spirit
to make the settings 'current working directory' contextual to the
same directory that CPPLINT.cfg is in.

This also makes execution consistent (picking up the "correct" settings)
regardless of the CWD when executing cpplint.py.

Example:

```
  echo 'root=..' >> /a/b/c/CPPLINT.cfg
  cd /
  cpplint.py /a/b/c/source_file.h
  # expects header guard of C_SOURCE_FILE_H_
```

However the old behavior would use '/../' = '/'
and incorrectly think the root was ``` 'A_B_C_SOURCE_FILE_H_'```.

### DIFF
--- a/cpplint/cpplint_unittest.py
+++ b/cpplint/cpplint_unittest.py
@@ -4218,6 +4218,8 @@ class CpplintTest(CpplintTestBase):
           error_collector.ResultList())
 
   def testBuildHeaderGuardWithRoot(self):
+    # note: Tested file paths must be real, otherwise
+    # the repository name lookup will fail.
     file_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
                              'cpplint_test_header.h')
     file_info = cpplint.FileInfo(file_path)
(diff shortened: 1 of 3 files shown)


iam:
Noticed root seems to be ignoring the directory of the CPPLINT.cfg so I uploaded this improvement:

@markmentovai

```
............................................................................................................................................................................
Ran 172 tests in 0.683s

OK
```

cpplint/cpplint_unittest.py
```diff
@@ -4238,9 +4240,30 @@ def testBuildHeaderGuardWithRoot(self):
...
+
+    # absolute directory
+    # (note that CPPLINT.cfg root=setting is always made absolute)
+    cpplint._root = os.path.join(os.path.dirname(os.path.abspath(__file__)))
+    self.assertEquals('CPPLINT_TEST_HEADER_H_',
+                      cpplint.GetHeaderGuardCPPVariable(file_path))
+
+    nested_file_path = os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                                    os.path.join('nested', 'cpplint_test_header.h'))
+    cpplint._root = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'nested')
```
markmentovai: 80 columns

markmentovai (changes requested):

iam:
Fixed line length. Also fixed tests (forgot to 'git add' a file).

@markmentovai

markmentovai (approved):

PR #293, opened 2017-11-10
cpplint: add --quiet flag to suppress all output when there's no errors

iam:
Example:
```
   cpplint.py --quiet <file-names>
```

Will now return with an exit code of ```0``` and return empty output if there
were no errors. This makes it particularly useful to be driven by a build system
such as makefiles or gradle.

In particular, these messages are now suppressed:
```
  Ignoring <filename>, excluded by CPPLINT.cfg, ...
  Done processing <filename>
  Total errors found: 0
```
If there were any errors, the above messages are printed nevertheless.

There is no behavior change if ```--quiet``` is not passed in.

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -57,6 +57,7 @@ _USAGE = """
 Syntax: cpplint.py [--verbose=#] [--output=vs7] [--filter=-x,+y,...]
                    [--counting=total|toplevel|detailed] [--root=subdir]
                    [--linelength=digits] [--headers=x,y,...]
+                   [--quiet]
         <file> [file] ...
 
   The style guidelines this tries to follow are those in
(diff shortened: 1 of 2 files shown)


iam:
@markmentovai 

Add new --quiet flag to suppress output:

```
$> ./cpplint/cpplint_unittest.py 
................................................................................................................................................................................
Ran 176 tests in 1.098s

OK
```

cpplint/cpplint.py
```diff
@@ -861,6 +865,7 @@ def __init__(self):
     self._filters_backup = self.filters[:]
     self.counting = 'total'  # In what way are we counting errors?
     self.errors_by_category = {}  # string to int dict storing error counts
+    self.quiet = False # Suppress non-error messagess?
```
markmentovai: The usual style (also used on the two lines above) is to put two spaces before the #.

iam: Solid catch. I fixed it. If only we had a pylint for our cpplint :).

markmentovai (approved):

PR #294, opened 2017-11-11
cpplint: Fix unittests when the styleguide directory is named otherwise

iam:
Don't hardcode the 'styleguide' directory in unit tests, instead
get it from the ```../..``` basename.

(It was failing when the repository directory was named differently, e.g. 'google-styleguide' directory, like we have in AOSP)


iam:
@markmentovai

markmentovai (approved):

PR #295, opened 2017-11-14
update readme.md

wangweitl81:


tonyruscoe:
I don’t think we want to make this change. Closing this without merging.

PR #306, opened 2017-12-15
Add java[x] imports as separate group in IntelliJ

textbook:
The STATIC###THIRD_PARTY_PACKAGE rule for customImportOrderRules and
the standardPackageRegExp mean that, previously, auto-formatting
IntelliJ imports would lead to Checkstyle failing, as the standard
packages are not grouped separately from third-party and same package
imports.

This creates a new group for java.* and javax.* imports, such that
IntelliJ can auto-format them correctly and pass validation.

### DIFF
--- a/intellij-java-google-style.xml
+++ b/intellij-java-google-style.xml
@@ -23,6 +23,9 @@
       <package name="" withSubpackages="true" static="true" />
       <emptyLine />
       <package name="" withSubpackages="true" static="false" />
+      <emptyLine />
+      <package name="java" withSubpackages="true" static="false" />
+      <package name="javax" withSubpackages="true" static="false" />
     </value>
   </option>
   <option name="RIGHT_MARGIN" value="100" />


ejona86:
java and javax are [not special cased](https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing) in the style guide any more. You should only have two groups: static and non-static imports.

textbook:
@ejona86 they are still in the CheckStyle version, though: https://github.com/checkstyle/checkstyle/blob/master/src/main/resources/google_checks.xml#L181. Perhaps I should be changing that end, rather than this one?

ejona86:
@textbook, yeah, CheckStyle should be updated. Although that's also interesting to me, because I see that same configuration in grpc/grpc-java and we've not had trouble. I guess that's because we're on an older version that is more permissive.

textbook:
Upgrading the Checkstyle used by the Maven plugin (see https://maven.apache.org/plugins/maven-checkstyle-plugin/examples/upgrading-checkstyle.html) seems to have restored the styleguide-compliant behaviour anyway.

PR #318, opened 2018-01-11
Fix typo in comment in Py style guide.

goody44:


goody44:
I signed it!

vapier:
`iff` is not a typo.  it's short for "if and only if".

https://en.wikipedia.org/wiki/If_and_only_if

Issue #320, opened 2018-01-16
Eclipse Java Line-Wrapping


JSlain:
In the [Java Style Guide](https://google.github.io/styleguide/javaguide.html#s4.5-line-wrapping), it's written that:

> [...] even code that would in fact fit within the column limit may be line-wrapped at the author's discretion.

However, the current settings for eclipse (`eclipse-java-google-style.xml`) makes the formatter forcing everything to be on the same line. There's nothing the developer can do to break it on multiple lines.

wallind:
this is still an issue to this date

PR #322, opened 2018-01-17
#320: That checks "Never join already wrapped lines"

JSlain:
This prevent removing hand-made line-wrapping when formatting code.
As specified in #320


JSlain:
I signed it!


/**
 * *Ghislain Nadeau*
 * Développeur Java
 * SSQ Groupe Financier
 * nadeau.ghislain@gmail.com
 */

JSlain:
Ping? Pong!

wallind:
Can we get this merged? It seems pretty clear [here](https://google.github.io/styleguide/javaguide.html#s4.5-line-wrapping) that this PR is a valid fix for an issue with the current style guide. Currently I have a source controlled local copy of the `eclipse-java-google-style.xml` in a project implementing this fix exactly and I would really like to swap to using the github hosted one here.

Unfortunately I can't do that till this line wrap fix is merged. I can validate it works if that makes a difference.

PR #324, opened 2018-01-29
Intellij INDENT_ON_CONTINUATION needs to be true

martin-linha:
As stated in the Style guide at paragraph `7.1.3 Block tags`:

> When a block tag doesn't fit on a single line, continuation lines are indented four (or more) spaces from the position of the @

Intellij's formatter needs to have "JD_INDENT_ON_CONTINUATION" flag set to true.


martin-linha:
I signed it!

PR #331, opened 2018-03-12
pylint does not find Python syntax errors or undefined names

cclauss:
__pylint__ does not find Python _syntax errors_ (like __print__ without the parens in Python 3) or _undefined names_ (like `unicode()` and `xrange()` in Python 3).  Use __python3 -m flake8 . --select=E9,F82 --show-source__ as a workaround.

### DIFF
--- a/pyguide.md
+++ b/pyguide.md
@@ -48,7 +48,10 @@ Catches easy-to-miss errors like typos, using-vars-before-assignment, etc.
 #### 2.1.3 Cons
 
 `pylint` isn't perfect. To take advantage of it, we'll need to sometimes: a)
-Write around it b) Suppress its warnings or c) Improve it.
+Write around it b) Suppress its warnings or c) Improve it.  `pylint` does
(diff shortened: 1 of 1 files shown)


gpshead:
Thanks, but this isn't actually true.  pylint must be running under the version of Python who's syntax you want it to check.  If you run pylint under python3 you will get an error such as:

```
E:  1, 0: Missing parentheses in call to 'print' (<string>, line 1) (syntax-error)
```

Similarly, if `from __future__ import print_function` is present it will flag that even under Python 2.

You are correct in noting that it doesn't flag name errors by default (especially for built-ins).  But I'd prefer we not mention this in the style guide or recommend flake8 as we do not use flake8 internally.

Regardless, we aren't ready to accept external contributions to our Python style guide.

cclauss:
@gpshead Understood.  Thanks for the tip on pylint.  I will put that to good use.

PR #339, opened 2018-03-31
Nit: s/template string/template literal/g

mathiasbynens:
The correct terminology is “template literal”. There is no such thing as a “template string”.

This patch leaves the `id` attribute value intact so that old links to this section don’t break.


vapier:
i think this has been fixed in the latest version now.  please check.

Issue #337, opened 2018-03-22
Issue with 16 bit representation of unicode characters


m-chaturvedi:
To reproduce:

```
#!/bin/bash

set -exuo pipefail

PYTHON_TAG_TO_CHECK=v2.7.4
PYTHON_UCS2="${PWD}/cpython/python_v2.7.10_ucs2"
PYTHON_UCS4="${PWD}/cpython/python_v2.7.10_ucs4"
NJOBS=16

git clone https://github.com/python/cpython.git
cd cpython
git checkout $PYTHON_TAG_TO_CHECK


mkdir ${PYTHON_UCS2}
mkdir ${PYTHON_UCS4}


test_with_cpp_lint() {
	local CMD="${1}/bin/python2.7 cpplint.py --filter=-build/header_guard,-legal/copyright initial_value_problem.h 2>&1"
	local output="$(eval ${CMD})"
	echo "${output}"
}

test_unicode_size() {
	if [ $(cat pyconfig.h | grep -cE "^\#define Py_UNICODE_SIZE $1$") -ne 1 ]; then
		echo "Py_UNICODE_SIZE ASSERTION FAILED"
		exit 1
			fi
}

check_assertion(){
    if [[ "$OUTPUT" != "$EXPECTED_OUTPUT" ]]; then
        echo "OUTPUT ASSERTION FAILED"
        exit
    fi
}


wget https://raw.githubusercontent.com/google/styleguide/gh-pages/cpplint/cpplint.py

# The motivation of the excercise (Line 18)
wget https://raw.githubusercontent.com/RobotLocomotion/drake/b6ff34893b83430c79043d0b6c29e00b8c04de52/systems/analysis/initial_value_problem.h

# The default is ucs2
# https://github.com/python/cpython/blob/1ce4e5bee6df476836f799456f2caf77cd13dc97/configure#L1473
./configure --prefix="${PYTHON_UCS2}" && make -j$NJOBS && make install
test_unicode_size 2

OUTPUT="$(test_with_cpp_lint $PYTHON_UCS2)"
EXPECTED_OUTPUT="$(cat << ENDOFTEXT
initial_value_problem.h:18:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
initial_value_problem.h:19:  Lines should be <= 80 characters long  [whitespace/line_length] [2]
Done processing initial_value_problem.h
Total errors found: 2
ENDOFTEXT
)"
check_assertion

# Checking with --enable-unicode=ucs4
[...]

PR #342, opened 2018-04-04
Fix Issue 337

m-chaturvedi:
[Similar cpython issue](https://bugs.python.org/issue4474)

Towards #337

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -51,6 +51,7 @@ import sre_compile
 import string
 import sys
 import unicodedata
+import sysconfig
 
 try:
   xrange          # Python 2
@@ -4291,6 +4292,16 @@ def GetLineWidth(line):
       if unicodedata.east_asian_width(uc) in ('W', 'F'):
         width += 2
       elif not unicodedata.combining(uc):
+        # Issue 337
+        # https://mail.python.org/pipermail/python-list/2012-August/628809.html
+        if (sys.version_info.major, sys.version_info.minor) <= (3, 2):
+          # https://github.com/python/cpython/blob/2.7/Include/unicodeobject.h#L81
+          is_wide_build = sysconfig.get_config_var("Py_UNICODE_SIZE") >= 4
+          # https://github.com/python/cpython/blob/2.7/Objects/unicodeobject.c#L564
+          is_low_surrogate = 0xDC00 <= ord(uc) <= 0xDFFF
+          if not is_wide_build and is_low_surrogate:
+            width -= 1
+          
         width += 1
     return width
   else:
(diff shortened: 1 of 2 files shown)


m-chaturvedi:
I signed it!

cpplint/cpplint.py
```diff
@@ -4291,6 +4292,15 @@ def GetLineWidth(line):
       if unicodedata.east_asian_width(uc) in ('W', 'F'):
         width += 2
       elif not unicodedata.combining(uc):
+        # Issue 337        
+        if (sys.version_info.major, sys.version_info.minor) == (2, 7) :
+          # https://github.com/python/cpython/blob/2.7/Include/unicodeobject.h#L81
+          is_wide_build = sysconfig.get_config_var("Py_UNICODE_SIZE") >= 4
+          # https://github.com/python/cpython/blob/2.7/Objects/unicodeobject.c#L564
+          is_low_surrogate = (0xDC00 <= ord(uc) and ord(uc) <= 0xDFFF)
```
gpshead: I believe this can be written as `0xDC00 <= ord(uc) <= 0xDFFF`

cpplint/cpplint.py
```diff
@@ -4291,6 +4292,15 @@ def GetLineWidth(line):
       if unicodedata.east_asian_width(uc) in ('W', 'F'):
         width += 2
       elif not unicodedata.combining(uc):
+        # Issue 337        
+        if (sys.version_info.major, sys.version_info.minor) == (2, 7) :
```
gpshead: I suggest writing this as `if sys.version_info <= (2,):` as this is not == 2.7 specific.  technically `<= (3,2)` makes sense but we're never going to support the long obsolete early 3.0-3.2 interpreters.

gpshead (commented):
is there a meaningful way to unittest this in cpplint_unittest.py?  (already there, or does one need to be added?)

Even a test that if run on both unicode width builds of cpython 2.7 interpreter would at least execute this code path to demonstrate the problem if this code weren't here would be good.

m-chaturvedi:
I don't follow this:

>Even a test that if run on both unicode width builds of cpython 2.7 interpreter would at least execute this code path to demonstrate the problem if this code weren't here would be good.

I added a test which will fail on UCS2 Python builds without the proposed change in `cpplint.py`.  Would that be sufficient?

cpplint/cpplint.py
m-chaturvedi: Thanks!

cpplint/cpplint.py
m-chaturvedi: Thanks!

m-chaturvedi:
Mac builds seem to be `ucs2`, btw, which is where this problem got exposed to us.

m-chaturvedi:
@gpshead Just a gentle reminder!  Thanks!

gpshead:
thanks, yes a test that will fail on such a built is good enough.

PR #343, opened 2018-04-09
Fixing Intellij style xml to keep prohibiting * imports.

romanoid:
Parameters likely moved in recent versions, keeping old params in place for people using older Intellij.


romanoid:
I signed it! 

(To be more precise, added myself to the google group with corporate CLA signed)

PR #349, opened 2018-04-15
Add Python3 Support of cpplint.py and cpplint_unittest.py

Johnson9009:
1. After alias unicode to str and replace iteritems, itervalues, cpplint.py can
   work well under Python3 if the linted file hasn't any problem. But if there
   is even only one violation of Google C++ Style Guide in linted file,
   cpplint.py will need to output error message, then exception like "TypeError:
   write() argument must be str, not bytes" will appear.
   This issue is caused by different supporting methods of Unicode between
   Python2 and Python3, in Python2 type "str" actually is bytes sequence, type
   "unicode" is used to handle Unicode string, but in Python3 type "str"
   definitely is Unicode sequence, so type "unicode" hasn't any value and be
   removed, type "bytes" is introduced to handle bytes sequence, so the type
   "bytes" of Python3 is equal to the type "str" of Python2.
   The type of argument of sys.stderr.write() is kept same between Python2 and
   Python3, all are "str", as mentioned above, actually the type of argument of
   sys.stderr.write() is changed totally. It receive a bytes sequence in Python2
   and a Unicode sequence in Python3. So in Python3 sys.stderr shouldn't be
   wrapped by a UTF-8 stream reader and writer of codecs.
2. Replace assert_ with assertTrue, assertEquals with assertEqual, because
   assert_ and assertEquals just alias of assertTrue and assertEqual, without
   this change below message will appear.
       DeprecationWarning: Please use assertEqual instead.
       DeprecationWarning: Please use assertTrue instead.

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -57,6 +57,23 @@ try:
   xrange          # Python 2
 except NameError:
   xrange = range  # Python 3
+  unicode = str
+  def iteritems(d):
+    return d.items()
+  def itervalues(d):
+    return d.values()
+else:
+  # Python 2
+  def iteritems(d):
+    return d.iteritems()
+  def itervalues(d):
+    return d.itervalues()
+  # Change stderr to write with replacement characters so we don't die
+  # if we try to print something containing non-ASCII characters.
+  sys.stderr = codecs.StreamReaderWriter(sys.stderr,
+                                         codecs.getreader('utf8'),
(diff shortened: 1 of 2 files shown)


Johnson9009:
I signed it!

tkruse:
Consider contributing to https://github.com/cpplint/cpplint

vapier:

PR #372, opened 2018-06-12
Link to the maintained pyguide.md, not old html.

gpshead:
pyguide.html is unmaintained and has been for ~5 years.  pyguide.md
is reasonably up to date and we have the ability to keep it up to
date from our internal guide.  Link to that instead.

I'd like to delete the old pyguide.html but a lot of things still
point directly at it so they at least have somewhere to land for
the time being.


eccentricon (approved):
All good! Thanks!

PR #373, opened 2018-06-17
Fix rendering for table in pyguide

leVirve:
There are missing spaces for the syntax of the table (matters in Github Markdown) that causes rendering error.

image


gpshead:
Thanks.  I see the cause of this on our end internally (copybara stripping a comment as it stripped some stuff out for the external version of the guide was leaving extra space).  I'm fixing that internally so that future updates won't repeat this problem.  In the interim, merging this PR.

PR #383, opened 2018-07-30
Add some design on the java.html

Gan9617:


tonyruscoe:
I don’t think we want to make these visual changes to all of the style guides which use this style sheet. Closing without merging.

PR #384, opened 2018-08-03
Update C++ style guide

pwnall:


pwnall:
@markmentovai Thank you very much for the quick review!

PR #392, opened 2018-09-10
Change the URL in the example of data.previousLink from /next to /previous

FujiZ:
I think /previous or /prev is more appropriate for previousLink.


FujiZ:
I signed it!

FujiZ:
I signed it!

PR #395, opened 2018-09-13
Correct "error" response JSON code snippet

jamesdaily:
Added missing double-quote within error.errors[].message `"File Not Found`, and applied linting as per https://jsoneditoronline.org


vapier:
your fix has been integrated via commit 8f97e24da04753c7a15eda6b02114a01ec3146f5.  I forgot to update this PR at the time.

PR #400, opened 2018-09-20
Added support for an 'access_indent' setting in CPPLINT.cfg

alulla:
It is now possible to override the access modifier indentation.
Previously this was hard-coded to 1


alulla:
I signed it!

PR #403, opened 2018-09-25
add wrap_long_lines in intellij-java-google-style.xml 

shenpengyan:
if one line code is too long, it is bad to watch code in screen, 

```
<option name="RIGHT_MARGIN" value="140" />
```

limit the one line code length, but we can not reformat it with ‘reformat code’ keymap when our code is longer than limit length.

with this set, we can reformat it 

```
<option name="WRAP_LONG_LINES" value="true" />
```


shenpengyan:
I signed it!

PR #408, opened 2018-10-08
cpplint: Make six-compatible for Python3

EricCousineau-TRI:
Upstream version of https://github.com/RobotLocomotion/styleguide/pull/13

This uses `six`, which may inject additional dependencies on downstream projects.
If so desired, I can rework it to be free of `six`.

---
This change is [Reviewable](https://reviewable.io/reviews/google/styleguide/408)

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -53,10 +53,9 @@ import sys
 import unicodedata
 import sysconfig
 
-try:
-  xrange          # Python 2
-except NameError:
-  xrange = range  # Python 3
+import six
+from six import text_type as unicode
+from six.moves import xrange
 
 
 _USAGE = """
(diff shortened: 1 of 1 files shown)


EricCousineau-TRI:
Hup, just saw #349, which accomplishes what this does without `six`.

Would be great to see that land. Will leave open until that PR lands or a maintainer closes this issue.

tkruse:
@EricCousineau-TRI : I don't think anyone at Google has time to look at PRs, let alone merge them.

Consider the fork at https://github.com/cpplint/cpplint (also published to Pypi) if you feel like contributing.

felker:
@tkruse is the fork active?

tkruse:
@felker : Somewhat. Don't expect much to move on it's own, but if you make a PR, and prod a bit, a new version will be released on pypi.

EricCousineau-TRI:
Gotcha, will file an issue on our side and see if we can update to use the fork for linting - thanks!

(And sorry for the late response!)

PR #409, opened 2018-11-07
Clearer example on python decorators

joseotoro:


gpshead:
Amusing, that appears to have happened from some errant internal -> external auto-editing (yuck).  we've got this wart fixed on our internal copy already.  I'll work on pushing out a fresh pystyle.md update.

gpshead:
fix pushed.

PR #413, opened 2018-11-16
Python: Fix confusion about class attributes

kannes:
Class attributes are NOT instance attributes. This example means instance attributes. My edit is as minimal as possible to fix that.


kannes:
Screw that, this is a trivial change. If needed, I hereby declare it to be in the public domain.

gpshead:
No worries.  For what it's worth, the Attributes: section in class docstrings applies to any form of attributes so we don't want to restrict this to _instance_ attributes here.

kannes:
D'oh, you are right.

PR #415, opened 2018-11-20
google/styleguide#412: Java: 4.1.2 example: Missing opening brace in …

### LINKED ISSUE #412: Java: 4.1.2 example: Missing opening brace in else if
The Java Style Guide states:

> **[4.1.1 Braces are used where optional](https://google.github.io/styleguide/javaguide.html#s4.1.1-braces-always-used)**
> Braces are used with if, else, for, do and while statements, even when the body is empty or contains only a single statement.
> 

But the example in the immediately following section contains:

> **[4.1.2 Nonempty blocks: K & R style](https://google.github.io/styleguide/javaguide.html#s4.1.2-blocks-k-r-style)**
> `    } else if (otherCondition()) {`

There is no opening brace `{` after the `else`! 

I think the example is wrong, it doesn't follow the rule stated in the previous section.

taurus227:
…else if

fixes #412

### DIFF
--- a/javaguide.html
+++ b/javaguide.html
@@ -294,10 +294,12 @@ return new MyClass() {
       } catch (ProblemException e) {
         recover();
       }
-    } else if (otherCondition()) {
-      somethingElse();
     } else {
-      lastThing();
+      if (otherCondition()) {
+        somethingElse();
+      } else {
+        lastThing();
+      }
     }
   }
 };


taurus227:
I signed it!

shicks:
This example is valid Java style (and syntax) as written, and the rule will not change to require a brace here.

taurus227:
@shicks
Please read the description of the issue #412, there is a contradiction between rule 4.1.1 and rule 4.1.2:

> image

I didn't say that it's not valid Java syntax. If all valid Java syntax was allowed, you wouldn't need a style guide.

PR #428, opened 2019-01-05
UTF-8 no BOM

veganaiZe:
Specify that no byte order mark should be used (when applicable) for source files.

### DIFF
--- a/javaguide.html
+++ b/javaguide.html
@@ -62,9 +62,9 @@ code. Optional formatting choices made in examples should not be enforced as rul
 (of which there is <a href="#s3.4.1-one-top-level-class">exactly one</a>), plus the
 <code>.java</code> extension.</p>
 
-<h3 id="s2.2-file-encoding">2.2 File encoding: UTF-8 no BOM</h3>
+<h3 id="s2.2-file-encoding">2.2 File encoding: UTF-8</h3>
 
-<p>Source files are encoded in <strong>UTF-8</strong>.</p>
+<p>Source files are encoded in <strong>UTF-8</strong> no BOM.</p>
 
 <h3 id="s2.3-special-characters">2.3 Special characters</h3>
 
--- a/javaguide.html
+++ b/javaguide.html
@@ -62,7 +62,7 @@ code. Optional formatting choices made in examples should not be enforced as rul
 (of which there is <a href="#s3.4.1-one-top-level-class">exactly one</a>), plus the
 <code>.java</code> extension.</p>
 
-<h3 id="s2.2-file-encoding">2.2 File encoding: UTF-8</h3>
+<h3 id="s2.2-file-encoding">2.2 File encoding: UTF-8 no BOM</h3>
 
 <p>Source files are encoded in <strong>UTF-8</strong>.</p>
 


veganaiZe:
I signed it!

veganaiZe:
The purpose of this commit is to avoid potential problems associated with BOMs, since they have been known to halt some script interpreters and compilers in their tracks.

Personally speaking, I have never experienced any problems by omitting the BOM.  But I write/code primarily in English.  If non-English writing/coding doesn't necessitate a BOM then it seems that the appropriate option would be to forgo its use, by default.

dimo414:
This doesn't really seem worth stipulating; the [BOM has no meaning for UTF-8](https://en.wikipedia.org/wiki/Byte_order_mark#UTF-8):

> The Unicode Standard permits the BOM in UTF-8,[3] but does not require or recommend its use.[4] Byte order has no meaning in UTF-8,[5] so its only use in UTF-8 is to signal at the start that the text stream is encoded in UTF-8, or that it was converted to UTF-8 from a stream that contained an optional BOM. The standard also does not recommend removing a BOM when it is there, so that round-tripping between encodings does not lose information, and so that code that relies on it continues to work.

What's the motivation for calling this out? Interpreters that don't handle BOMs sound like they aren't properly implementing the UTF-8 spec and should be fixed.

veganaiZe:
The reason is that back-compatibility with ASCII is a major benefit of UTF-8.

For instance a Windows `.bat`/batch file saved as UTF-8 + BOM will crash when run because the command interpreter doesn't understand the BOM (ie. `∩╗┐`) prefix -- It's just garbage, which also displays in some text editors; which could easily confuse one as to whether to let this _mysterious garbage_ persist, or should they remove it?

The following Unicode laced `.bat` file, saved as UTF-8, will work properly only when saved _**without**_ a BOM:
```
echo "☺ Ł" >textfile.txt
```

As the standard itself states, it neither requires nor recommends the use of a BOM, with UTF-8.

The so-called "round-tripping" scenario is really only appropriate when dealing with (pre-existing / external) text files, input into a program --not to the (Google) source code to be executed.

So ideally one shouldn't erase existing BOMs, willy-nilly, because they may be processed by another system which, against the standard's recommendation, expects their presence.  However, when fresh source files are created, the BOM is probably best avoided, in accordance with the Unicode/UTF-8 recommendation.

It follows the programming principle: "Be liberal in what you accept and conservative in what you send."

dimo414:
I don't think this is worth adding. As you say, the UTF-8 spec already discourages using a BOM when feasible, and I don't see any reason why it being a Java source file introduces further complexity that merits calling this out in the style guide. I'm not certain I've ever even seen a UTF-8 file with a BOM...

veganaiZe:
If it's worth stating that the file should be in UTF-8 format then it's worth mentioning the avoidance of a BOM.

But that's just my little ol' opinion.

veganaiZe:
utf-8_bom
[ What if my editor is dumb and pre-checks the BOM box, for UTF-8 files, or just includes it without giving me any option? ]

dimo414:
It's not the goal of the Java Style Guide to address all possible configuration issues; we require Java source code to be UTF-8 for numerous reasons, but the presence or absence of a BOM is not something we need to make a ruling on. Leave it off when possible, as the UTF-8 spec already says.

veganaiZe:
So as long as everyone reads through at least the first 39 pages of the Unicode standard then there shouldn't be any problem.  Why would anyone even consider using `.BAT` files to automate Java code compilation or execution?  I'm such an idiot.  I'm so sorry for wasting your precious time with a totally realistic scenario; and providing examples to support it.

Meanwhile, you've provided zero evidence to support your claim, which is based solely on your personal experience.  The Wikipedia snippet, a somewhat pathetic and non-authoritative source, in all honesty, actually supports my argument.

If I ever make a contribution to Google I will be sure to include that BOM.

It's no wonder that even Google's own engineers could care less about the company's style guides.

And here I'd thought that Google workers were above mansplaining.  Stupid me.

dimo414:
Hi veganaize, first, let me say thank you for sending this PR; I should have done so in an earlier comment. Rejecting the PR is in no way a statement about you, nor your qualifications. Although it's true that Google's public style guides are under-curated (a situation we all lament, but no one has stepped up to address) we care very deeply about the contents of our style guides, and therefore are very conservative in what we add to it. My intent here was to bring closure to this issue, since it had unfortunately sat unaddressed for some time.

You are absolutely right that there are situations where a BOM can matter, and it is reasonable to discourage their use. In our experience at Google this has not been a major issue (admittedly, most internal development happens on Linux, so issues related to `.bat` scripting are indeed something we don't typically run into). Looking into historical internal discussions around BOMs affecting Google Java developers I can find very little, and nothing related to BOMs in the `.java` source files themselves. None of which invalidates your concern, but in our opinion it doesn't rise to a level of severity to merit calling out in the Style Guide. Many best practices are intentionally not included in the guide simply because we don't want to make a ruling we don't have to. We trust teams to identify their own best practices where we don't make stipulations.

Of note, this doesn't appear to be a Java-specific issue, and I notice only the HTML/CSS guide makes any mention of it. Perhaps we should consistently discourage BOMs across all our style guides, but I don't believe it's necessary to make the Java guide an outlier in this regard.

> If I ever make a contribution to Google I will be sure to include that BOM.

I would ask that you do so with good-faith, but *please do* include a BOM if you believe it is relevant. We put a lot of faith in our tooling, and count on developers both internally and externally to help catch places where our tooling falls short. If a BOM breaks something we rely on, we will want to fix it. You might even [get paid](https://www.google.com/about/appsecurity/reward-program/) if you are able to find an exploit related to mishandling of BOMs.

> And here I'd thought that Google workers were above mansplaining. Stupid me.

One of the things I have deeply, deeply valued about my time at Google has been the candid and open discussions around issues of identity, gender inequality, and respect. We continue to fall short in many ways, but I am proud of the effort my peers have made to create a culture that is welcoming and inclusive. I don't believe anything I've said could fairly be labeled mansplaining, but I am sorry that I was curt. Again, thank you for raising this issue.

PR #432, opened 2019-02-04
[cpplint] detection of broken std::regex in libstdc++

alexeyserbin:
Added detection of broken std::regex and friends in older
versions of libstdc++ (< 4.9) into the cpplint tool.  For details,
see [1].

The motivation for this change is the fact that some projects still
need to be compiled and linked against older libstdc++ at still
supported Linux distros such as RHEL/CentOS 7, Ubuntu 14.04 and SLES 12,
but using C++11 std::regex functionality could lead to unexpected crash
if the code is compiled with C++11-compatible compiler and linked aginst
libstdc++ of versions prior to 4.9.

References:
  [1] https://gcc.gnu.org/onlinedocs/gcc-4.8.2/libstdc++/manual/manual/status.html#status.iso.2011
      section 28 'Regular Expressions'

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -235,6 +235,7 @@ _ERROR_CATEGORIES = [
     'readability/todo',
     'readability/utf8',
     'runtime/arrays',
+    'runtime/broken_libstdcpp_regex',
     'runtime/casting',
     'runtime/explicit',
     'runtime/int',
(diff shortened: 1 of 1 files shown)


alexeyserbin:
I signed the CLA.

tkruse:
Consider contributing to the fork at https://github.com/cpplint/cpplint

vapier:

PR #438, opened 2019-02-20
Keyword 'override' can be now enforced by gcc or clang-tidy.

jankratochvil:


vapier:
I don't think the style guide discusses compiler options, or specific compilers in general.  So I don't think we'd want to refer to them here.

PR #444, opened 2019-04-05
shell: fix ShellCheck error SC2145 in err function

mpoullet:
For more information:
  https://www.shellcheck.net/wiki/SC2145 -- Argument mixes string and array. ...


IsaacG:
Thank you for your PR!
[The code](http://google.github.io/styleguide/shellguide.html#stdout-vs-stderr) has since been updated to use `$*`

PR #448, opened 2019-04-17
update jQuery version in example

coliff:
I know it's only an example, but best to have the most recent version here which doesn't have a known security vulnerability.

REF:
- https://snyk.io/vuln/npm:jquery
- https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/


tonyruscoe:
LGTM. Make sense. Thanks for contributing, @coliff!

PR #449, opened 2019-04-18
Updated shebang recommendation for python files

ryanrichholt:
This was fixed in #149 but it looks like the change got lost somehow when the switching from the .html to .md file


ryanrichholt:
I signed it!

vapier:
commit 5120fd6b14e3135745668dc8eeb1c5581d3f4d9b (Dec 2020) included this fix

PR #453, opened 2019-06-03
Update jQuery to latest version in example

coliff:


tonyruscoe:
I know we updated the jQuery version in #448 because the version we’d been using in the example had a known security vulnerability, but I don’t think we need to update it every time a new version is available.

If having an old version of jQuery in the example is problematic, I’d prefer to replace it with a fictional script, but we don’t intend the style guide to provide “copy and paste” examples, so I don’t anticipate it causing any issues.

PR #454, opened 2019-06-05
changed @ to pie, as in 'pie' notation

simbamford:


gpshead:
Thanks!  I'm pulling these in internally and will include them in a future push.

I've never heard @ called pie notation so I'm letting the internet decide: https://twitter.com/gpshead/status/1136373118044217344

simbamford:
> Thanks! I'm pulling these in internally and will include them in a future push.
> 
> I've never heard @ called pie notation so I'm letting the internet decide: https://twitter.com/gpshead/status/1136373118044217344

Okay, no problem. You see it used here, for example: 

https://wiki.python.org/moin/PythonDecorators#A1._pie_decorator_syntax

By the way, I thought you might like to see this:

http://www.sim.me.uk/cheesecake/

Cheers
Sim

vapier:
typos and such have been integrated.  the topic of pie is still open.

I think the history is clear in this regard -- it was called "pie syntax" as a tongue-in-cheek reference.
> https://peps.python.org/pep-0318/#background
> Barry Warsaw named this the ‘pie-decorator’ syntax, in honor of the Pie-thon Parrot shootout which occurred around the same time as the decorator syntax, and because the @ looks a little like a pie.

> https://wiki.python.org/moin/PythonDecorators#A1._pie_decorator_syntax

all that said, I don't think the name has really stuck.  everyone knows "decorator", and the Python manual itself doesn't mention pie.
* https://docs.python.org/2/glossary.html#term-decorator
* https://docs.python.org/3/glossary.html#term-decorator

so I'm inclined to not have the guide refer to it as "pie".

PR #456, opened 2019-06-09
Make indentation consistent in pyguide.md

funcodeio:


funcodeio:
Since the guide's indentation is 4, it is good to have 4-space indentation in all examples through out the document. There were 2-space examples which made me confused.

gpshead:
Thanks.  Agreed this looks odd to people.

For what it's worth, the reason our guide is a bit messy this way is that we use two space indentation for our internal style but don't recommend that to the rest of the world which standardizes on four.  We currently do not have an appropriate transformation setup to fix all of this in the examples as we publish the external version of the guide from our internal master.

I'm leaving this PR open assigned to me to deal with that one way or another.

vapier:
so your point is def correct that the examples should be using 4 spaces.  i think it might be a pain for us to handle internally though :(.  we have an internal guide that we sanitize to produce the public guide.

vapier:
ah, it looks like we do exactly that internally in some examples :(.  we duplicate some code examples between internal (2 space) and external (4 space), but we didn't do it for all of them.  oof.

PR #466, opened 2019-09-03
Remove blanks before semicolons

thiemonagel:
All the code in the doc has semicolons attached to the preceding token
without a blank, except in three places. Bringing these three places in
line with the rest.


IsaacG:
Thank you for your PR!
I believe the guide no longer has any <space><single-semicolon>s in the current iteration.

PR #467, opened 2019-09-04
Setting the 'New line after (' property to correctly match the code styleguide

ws141:
This PR fixes the issue of the property 'New line after (' in the 'Method declaration parameters' section for Java not being set to true when importing the `intellij-java-google-style.xml`. See screenshot below.

Screenshot 2019-09-04 at 07 52 05

When the property isn't set the code is formatted in a way that doesn't pass the formatting check by `fmt-maven-plugin`.

Please also be aware that there seems to be a bug in IDEA currently which doesn't apply this property correctly if its imported with the same code style schema name multiple times so please use a new name when testing. For some reason IDEA keeps the latest saved state of the property in that code style schema.


ws141:
@shicks could you have a look at this PR? I noticed you have approved similar PR:s before.

PR #474, opened 2019-10-01
Favicons have been added for all html pages

sbenstewart:


vapier:
should be handled in latest tree now, thanks

PR #480, opened 2019-10-09
Update htmlcssguide.html

vanshika703:


tonyruscoe:
I don’t think we want to make this change. Closing without merging.

PR #487, opened 2019-11-18
Add support for c++17 tuple destructuring

allight:
C++17 adds support for tuple destructuring. This allow one to write
code such as:

```
std::pair<int, int> span = getSpan();
auto [start, end] = span;

// Use start as span.first and end as span.second
```

This makes cpplint recognize and allow a space before the '[' in this
situation.

This is a purposeful divergence from the internal version where the
entire whitespace/braces category was removed. It was decided to leave
the checks in since this is sometimes used without other formatting
tools.

Test: manual


allight:
@tituswinters @markmentovai

markmentovai (approved):

PR #490, opened 2019-12-22
Update pyguide.md

brlala:
added delay type checking syntax for python 3.7 and above for 3.19.13 Conditional Imports


brlala:
@googlebot I signed it!

gpshead:
I won't merge this PR directly as we maintain the source of truth for the Python style guide internally and push it outwards, but we're heading in this direction regardless: Once we're fully on 3.7+ (soon), we'll be recommending `from __future__ import annotations` everywhere and will update the guide with wording similar to this.

brlala:
Thanks appreciate that you take the time to reply!

PR #491, opened 2019-12-22
Delayed type checking for Type Hinting python 3.7+

brlala:
added delay type checking syntax for python 3.7 and above for 3.19.13 Conditional Imports


brlala:
Closed with reference to #490

PR #508, opened 2020-01-26
some typing advice

shendeguize:
change space dismatching btween 4 spaces and 2 spaces,
also move yes and no outside code blocks so matching with line225

### DIFF
--- a/pyguide.md
+++ b/pyguide.md
@@ -668,7 +668,7 @@ Use generators as needed.
 <a id="291-definition"></a>
 
 <a id="generators-definition"></a>
-#### 2.9 Definition 
+#### 2.9.1 Definition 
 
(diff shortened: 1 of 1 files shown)


shendeguize:
And there are some more confusing usage of two spaces and four spaces in python examples.
Also some "yes:" and "no" are in ```python```block some are outside, which is not continous.
I tried to edit all these parts but finally was not sure which type to choose, so just PR an example here. Looking forward to further reply.

shendeguize:
line671 should be 2.9.1

shendeguize:
also at line799, it would be better with yes: and no:

crypt096 (approved):

tkohli (approved):

ProKashif (approved):

PR #522, opened 2020-02-17
Add an HTML META tag to redirect to the new shell guide.

IsaacG:


dimo414:
Any idea if we can add a [rel=canonical](https://support.google.com/webmasters/answer/139066) tag to this page pointing at the new guide? The old XML guide still shows up at the top of searches for \[google shell style\] and similar - rel=canonical should fix that.

IsaacG:
I don't see why not...

IsaacG:
https://github.com/google/styleguide/pull/546

Want to review it for me?

PR #528, opened 2020-02-26
Add python 3 support to cpplint and cpplint_unittest

m-chaturvedi:


cpplint/cpplint_unittest.py
```diff
@@ -37,6 +37,7 @@
 import os
 import random
 import re
+import six
```
vapier: this is a third party import, so it should be after unittest but before cpplint

cpplint/cpplint_unittest.py
```diff
@@ -3071,7 +3072,8 @@ def DoTest(self, raw_bytes, has_invalid_utf8):
       error_collector = ErrorCollector(self.assert_)
       cpplint.ProcessFileData(
           'foo.cc', 'cc',
-          unicode(raw_bytes, 'utf8', 'replace').split('\n'),
+          six.ensure_text(raw_bytes, encoding='utf8',
+            errors='replace').split('\n'),
```
vapier: indent is off -- it should align with ( above

that said, you don't need six.  just write `raw_bytes.decode('utf-8', 'replace')`.

cpplint/cpplint_unittest.py
```diff
@@ -3104,7 +3106,7 @@ def testBadCharacters(self):
     cpplint.ProcessFileData(
         'nul_utf8.cc', 'cc',
         ['// Copyright 2014 Your Company.',
-         unicode('\xe9x\0', 'utf8', 'replace'), ''],
+          six.ensure_text(b'\xe9x\0', encoding='utf8', errors='replace'), ''],
```
vapier: indent is off -- should be aligned with [ above

also as above, you don't need `six`.  call `.decode()` on the byte directly.

cpplint/cpplint_unittest.py
```diff
@@ -5744,9 +5748,10 @@ def testQuietWithErrors(self):
 
   def testNonQuietWithoutErrors(self):
     # This will succeed. We filtered out all the known errors for that file.
-    (return_code, output) = self._runCppLint('--filter=' +
+    (return_code, output_bytes) = self._runCppLint('--filter=' +
                                                 '-legal/copyright,' +
```
vapier: indent is off now

cpplint/cpplint_unittest.py
```diff
@@ -5758,10 +5763,11 @@ def testNonQuietWithoutErrors(self):
 
   def testQuietWithoutErrors(self):
     # This will succeed. We filtered out all the known errors for that file.
-    (return_code, output) = self._runCppLint('--quiet',
+    (return_code, output_bytes) = self._runCppLint('--quiet',
                                              '--filter=' +
```
vapier: indent is off now

vapier (changes requested):

cpplint/cpplint_unittest.py
m-chaturvedi: Removed.

cpplint/cpplint_unittest.py
m-chaturvedi: OK.

cpplint/cpplint_unittest.py
m-chaturvedi: OK.

cpplint/cpplint_unittest.py
m-chaturvedi: OK.

cpplint/cpplint_unittest.py
m-chaturvedi: OK.

m-chaturvedi:
@vapier I was making these changes to the tests for our internal cpplint which already has support for python3.
I realized that I need to change the corresponding `cpplint.py` here as well.

We need to substitute `assertEquals` with `assertEqual` and `assert_`  with `assertTrue` to remove the deprecation warnings (will make another PR once / if this is accepted): 
https://docs.python.org/3.3/library/unittest.html#deprecated-aliases

vapier:
@markmentovai what's the story with cpplint in this repo ?  is this the master copy ?  or is it in google3 ?  or is it cpplint/cpplint ?

markmentovai:
[cpplint here](https://github.com/google/styleguide/tree/gh-pages/cpplint) was maintained by @eglaysher while he was at Google, and hasn’t seen much action since.

I didn’t know anything about [cpplint/cpplint](https://github.com/cpplint/cpplint) until moments ago.

gpshead:
I have pointed others who have come along at the https://github.com/cpplint/cpplint community maintained one as well due to lack of traction in getting any of us to maintain the version in this repo.

*Googlers*: internally our `devtools/cpplint/` version has been updated to use Python 3.  I believe all that needs is a copybara config setup to export it+test to this repo from time to time.  For an example of how to set that up, code search internally for `config/pylint_only/copy.bara.sky`.

Alternatively, if we're not going to maintain the version here I suggest deleting it and officially pointing external users at the aformentioned cpplint project.

vapier:
if we do have an internal copy to publish from, then publishing it here would be helpful.  i think i've had it come up with other public repos that we wanted updated cpplint scripts.

markmentovai:
It’s just a question of ownership. If there are internal owners for cpplint, it’d be best for them to be involved in the process. But even if we can’t get that, if someone’s willing to do it best-effort (as Elliot had been doing) or even as a one-off, it’d be an improvement.

raselranajoy688 (approved):

tkruse:
See https://github.com/google/styleguide/pull/837

vapier:

PR #529, opened 2020-02-26
Remove redundant instructions about function location

mattalexander-pantheon:


mattalexander-pantheon:
@googlebot I signed it!

vapier:
thanks for the suggestion.  i've posted an internal CL that merges the two paragraphs in a different way, but does what you propose.

unfortunately we can't accept PRs to the shellguide because this is a generated output file :(.  you can just submit issues instead as that should be less overhead/wasted effort for you.

mattalexander-pantheon:
Got it. Then I think the "Improve this page" link at the bottom of the page should link to the Issues page, not to the file edit page. @vapier

vapier:
hrm, i'm not sure we have control over that :(.  that seems to be part of the site template and not in the shell guide itself.

i'll ask around to see what the expectations are and if we can adjust the template for all pages.

sorry for the confusion.

vapier:
we've updated the site now so that link no longer appears.

feel free to file issues for us though when you notice problems like this.

Issue #505, opened 2020-01-23
Python - multi-Args / Returns indentation formatting


ColinKennedy:
The style guide for Python shows that arguments go on the first line and each line after are indented. This has led to some rather strange readability issues.

Such as this:

```python
        """Keep track of the current Rez package and the user's settings.

        Args:
            package: The current Rez package that is being checked.
            processed_packages: Other install Rez packages that have
                already been processed. These packages should be
                pre-validated and are meant only for caching. Default is None.
            allow_vimgrep: Some user input to track that specifies if they want row/column data printed.
            verbose: Some user input to track that specifies
                if they want summary information or everything to be
                printed.

        """
```

And this (pre-type annotations)

```python
def foo(something=[(8, 1.0), (-1, 11.0), (3, 4, 6, 7)], another_argument=frozenset(), more=None):
    """Do something.

    Args:
        something (iter[tuple[int, float] or tuple[int]], optional): My information.
        another_argument (set[str], optional): Lorem ipsum dolor sit amet, consectetur 
            adipiscing elit. Ut erat urna, finibus non tincidunt scelerisque, commodo eget ante. 
        more (None, optional): Aliquam vel justo convallis est sollicitudin egestas et quis arcu. 
           Nam ut mauris luctus, elementum dolor eu, placerat turpis. Ut non consequat augue, 
           sit amet dapibus arcu. Vivamus mattis dui vel mi condimentum, pellente
    """
```

Is it a requirement that the description start on the same line or may it start on the line below? (example)

```python
        """Keep track of the current Rez package and the user's settings.

        Args:
            package:
                The current Rez package that is being checked.
            processed_packages:
                Other install Rez packages that have already been
                processed. These packages should be pre-validated and
                are meant only for caching. Default is None.
            allow_vimgrep:
                Some user input to track that specifies if they want
                row/column data printed.
            verbose:
                Some user input to track that specifies if they want
                summary information or everything to be printed.

        """
```

What are the common strategies used when dealing with docstrings that mix single and multi-line descriptions in the same function?

gpshead:
We've got a custom pylint GoogleDocstringChecker internally that verifies the our style of docstring format.  My internal response to this kind of question is generally to defer to whatever our lint checker implementation asks for when it comes to details like this.  Running your final example through that, it likes it.  Which is a strong indicator that we allow that nice readable form. :)

[...]

PR #537, opened 2020-03-06
Added a clause for "indented-line" blocks for the Python style guide

ColinKennedy:
There's a discussion going on in pydocstyle (Reference: https://github.com/PyCQA/pydocstyle/issues/449) about what whitespace options are allowed under google's style guide. In [this issue submitted to this repository](https://github.com/google/styleguide/issues/505), it was mentioned that using indented lines is allowed by Google's current linters.

Would it be alright to add an explicit example?

### DIFF
--- a/pyguide.md
+++ b/pyguide.md
@@ -1982,15 +1982,39 @@ aptly described using a one-line docstring.
 
 <a id="doc-function-args"></a>
 [*Args:*](#doc-function-args)
-:   List each parameter by name. A description should follow the name, and be
-    separated by a colon and a space. If the description is too long to fit on a
-    single 80-character line, use a hanging indent of 2 or 4 spaces (be
-    consistent with the rest of the file).
+:   List each parameter by name. A description should follow the name on the
+    same line and be separated by a colon and a space. A description can also
+    follow on the next line. If the description is too long to fit on a single
+    80-character line, use a hanging indent of 2 or 4 spaces (be consistent with
+    the rest of the file).
 
     The description should include required type(s) if the code does not contain
     a corresponding type annotation. If a function accepts `*foo` (variable
     length argument lists) and/or `**bar` (arbitrary keyword arguments), they
     should be listed as `*foo` and `**bar`.
+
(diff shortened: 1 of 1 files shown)


rmorshea:
@gpshead any progress on this?

pyguide.md
```diff
@@ -1982,16 +1982,40 @@ aptly described using a one-line docstring.
...
     length argument lists) and/or `**bar` (arbitrary keyword arguments), they
     should be listed as `*foo` and `**bar`.
 
+    Both examples below are valid descriptions.
+
+```python
+def print_data(verbose):
+    """Prints information about the user's environment.
+
+    Args:
```
sambhav, line 2002: Can we also clarify what to do in case of a mix of both the styles? And add an example for that as well?

ColinKennedy:
@gpshead could you please update on this review? Or give details on how to advance this PR?

pyguide.md
rmorshea: If I recall, there'd been some discussion [here](https://github.com/PyCQA/pydocstyle/issues/449) in favor of allowing mixed styles across a code base, but not within a single docstring. That may be over prescriptive though and should probably be decided by someone at google. cc: @gpshead

gpshead:
I just mailed an internal change that should address this.  I'll make sure the the external guide is synced after that goes in.

pyguide.md
gpshead: in general i favor not being over prescriptive about that.  the main consistency thing is the indentation level (which is already mentioned).  but trying to enumerate a matrix of possible argument or exception and their descriptions style mixes for things would explode meaninglessly and become an enemy of documentation. :)

ColinKennedy: As long as "hanging indent" style is allowed by the standard, I'm happy for maintainers of tools to decide whether or not to enforce consistency. Avoiding being over prescriptive and having the option to pick the right look for the right doc would be a great thing.

gpshead:
An update including my internal change addressing this has been pushed.

PR #542, opened 2020-04-12
Fix argument name

deepyaman:
The above examples use the same variable as an input to `mybinary`; it doesn't make sense/seems like an oversight to pass an undeclared `get_arguments` in this case.


deepyaman:
@googlebot I signed it!

vapier:
the current style guide is correct (it's trying to show two unrelated commands that are bad), but i can see how it can be confusing, especially with the other examples here being related.  we'll try to clarify it internally and push out a new version.  thanks for highlighting it.

PR #543, opened 2020-04-14
C++: Do not use [=] or [=,this] to capture 'this'

pdillinger:
Summary: C++20 deprecates [=] capturing 'this'
(http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0806r2.html)
but the recommended alternative [=,this] is not standards-legal prior to
C++20 (though it is widely accepted *with warning*).

For portability (and readability), style should forbid using
[=] or [=,this] to capture 'this'. Any by-value captures must be
explicitly listed if also capturing 'this', though 'this' can still
be implicitly default-captured by reference, as in
[&, some_var_to_copy].

Test Plan: documentation only

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -3479,7 +3479,8 @@ array(T, U...) -&gt; std::array&lt;T, 1 + sizeof...(U)&gt;;
 <h3 id="Lambda_expressions">Lambda expressions</h3>
 
 <p>Use lambda expressions where appropriate. Prefer explicit captures
-when the lambda will escape the current scope.</p>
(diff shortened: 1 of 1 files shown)


tituswinters (commented):
Thanks for your PR. Lemme run this past the other owners of the style guide and get back to you.

My guess is this may be more words than we want to devote on the topic in this document: supplemental material like the TotW series (abseil.io/tips) may be better for the more-in-depth guidance parts of this.

I'll get back to you in a couple days either way.

PR #546, opened 2020-04-30
Add a canonical link to update the URL for the Shell guide

IsaacG:


dimo414:
LGTM, thanks!

dimo414:
It looks like Google has re-indexed the page, but is still showing it 😞 [This post](https://stackoverflow.com/q/5086894/113632) is far from canonical, but it's possible the Google Crawler doesn't actually handle XSL properly, which would be a bit ironic.

IsaacG:
I wonder if it would handle a <rel=canonical> XML tag...

dimo414:
I'd be a bit surprised if it did, but I'm not sure.

PR #553, opened 2020-05-20
Update C++ styleguide

zetafunction:
Major changes:
- discourage const references when the input parameter must outlive the
  call
- drop ban on mutable references
- allow rvalue-qualified methods in certain situations
- allow C-style cast when casting to void
- allow designated initializers in C++20 compliant form
- disallow macro-style naming for enumerator constant for new code
- disallow spaces inside parentheses for conditionals

The change itself is largely automated, with a few manual adjustments:
- manually restore the TOC placeholder
- removed trailing whitespace

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -10,9 +10,6 @@
 <body onload="initStyleGuide();">
 <div id="content">
 <h1>Google C++ Style Guide</h1>
-<div class="horizontal_toc" id="tocDiv"></div>
-<div class="main_body">
-
 <h2 id="Background" class="ignoreLink">Background</h2>
 
 <p>C++ is one of the main development languages  used by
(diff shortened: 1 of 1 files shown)


zetafunction:
@tituswinters please take a look when you can.

I've tried to summarize the more major changes in the commit description, since there's a lot of changes that just add a comma after "i.e." and "e.g.".

pwnall:
I looked through the PR. It is suitable for external consumption, and the summary is accurate.

I consider the following change to also be major, in the sense that I'd like to be aware of while reviewing.

* disallow postfix form of the ++ and -- operators in most cases (only allowed when the postfix semantics are needed)

cppguide.html
```diff
@@ -2201,23 +2143,33 @@ <h3 id="Rvalue_references">Rvalue References</h3>
...
-<p>You may use forwarding references in conjunction with <code>
-<a href="http://en.cppreference.com/w/cpp/utility/forward">std::forward</a></code>,
-to support perfect forwarding.</p>
+<p>Do not use rvalue references (or apply the <code>&amp;&amp;</code>
+qualifier to methods), except as follows:</p>
+<ul>
+  <li>You may use them to define move constructors and move assignment
+  operators (as described in
+  <a href="#Copyable_Movable_Types">Copyable and Movable Types</a>). See the
+  <a href="primer#copying_moving">C++ Primer</a> for more information about
```
tituswinters: The primer is internal only, we should trim this sentence / link.

zetafunction: Good catch—fixed!

zetafunction:
@tituswinters when you get a chance, PTAL. I think it should be ready for another look.

Sorry, I don't use GH a lot so not sure exactly how reviews on PRs work :)

tituswinters (approved):

tituswinters:
No worries, I was just on vacation a few days. Thanks!

PR #555, opened 2020-05-27
[C++ styleguide] Restore table of contents

zetafunction:
This was mistakenly dropped when updating the styleguide from the
internal C++ styleguide.


zetafunction:
@tituswinters Please take a look when you have the chance, thanks!

tituswinters (approved):

PR #562, opened 2020-06-19
Remove blacklist/whitelist from cpplint.py

nealrichardson:
In https://github.com/apache/arrow/pull/7484 we recently removed instances of "blacklist" and "whitelist" from our repository. This includes a copy of cpplint.py, so we wanted to submit those changes upstream.


nealrichardson:
@googlebot I signed it!

PR #570, opened 2020-07-29
Update language to comply with Android's inclusive language guidance

allight:
See https://source.android.com/setup/contribute/respectful-code for
reference

cpplint.py is copied into AOSP from this github version so update it
here first.

Bug: 161896447
Test: None (comments only)


allight:
@tituswinters @markmentovai PTAL

allight:
@vapier PTAL

gpshead:
Do realize that Google has been negligent on maintaining cpplint as found in this repository.  A community maintained fork exists in https://github.com/cpplint/cpplint.

allight:
@gpshead do you suggest we switch over to that version?

PR #577, opened 2020-09-24
Update C++ styleguide

zetafunction:
- add section for inclusive language.
- move include-what-you-use to a new section.
- clarify usage of absl::optional for by-value vs by-reference params.
- document usage of `` as a code-indexing hint
- note that enclosing identifiers in `` can help code indexers.

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -247,14 +247,37 @@ guard:</p>
 
 
 
+<h3 id="Include_What_You_Use">Include What You Use</h3>
+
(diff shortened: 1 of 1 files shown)


zetafunction:
@googlebot I consent.

zetafunction:
@tituswinters Please take a look when you have a chance. Thanks!

In the future, I'm going to try to do these merges regularly every quarter.

tituswinters (approved):

PR #581, opened 2020-10-29
Removed the non working link on Build System

SayarB:
The link on Build system in pyguide was not working  so i removed the link to avoid miscommunication.


gpshead:
thanks.  i'll make this fix internally and push it out.  that's a link to a section we omit from the public version of the guide where it doesn't make much sense to have.

PR #584, opened 2020-11-11
Update C++ styleguide

zetafunction:
- Guidance for return values (prefer return by value, return by
  reference, then return by pointer in that order)
- Mandate curly braces around the body of a control statement,
  providing one exception for historical reasons (`if` with no `else`,
  and `if` + body fits completely in a single line or the `if` fits in
  one line and the body fits in another separate line).

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -1693,8 +1693,8 @@ earlier.</p>
 <code>protected:</code>, then <code>private:</code>.  Omit
 sections that would be empty.</p>
 
-<p>Within each section, generally prefer grouping similar
-kinds of declarations together, and generally prefer the
+<p>Within each section, prefer grouping similar
+kinds of declarations together, and prefer the
 following order: types (including <code>typedef</code>,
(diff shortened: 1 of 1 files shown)


zetafunction:
@tituswinters Please take a look when you have a chance. Thanks!

cppguide.html
```diff
@@ -1719,6 +1719,9 @@ <h3 id="Inputs_and_Outputs">Inputs and Outputs</h3>
 improve readability, and often provide the same or better
 performance.</p>
 
+<p>Prefer to return by value or, failing that, return by reference.
+Avoid returning a pointer unless it can be null.</p>
```
wp-seth, line 1723:
```suggestion
Avoid returning a pointer unless you take care of the case if it is null.</p>
```
otherwise it could be misunderstood.

wp-seth (commented):
just a small note. all in all i welcome those corrections/updates/simplifications/precisions.

tituswinters (approved):

PR #597, opened 2020-11-11
Add missing function body

manoharreddyporeddy:
Add missing function body


shicks:
Merging this internally.  Thanks for pointing this out, it should roll out with the next update.

PR #598, opened 2020-11-11
Add missing "function" keyword

manoharreddyporeddy:
Add missing "function" keyword


shicks:
Fixing this internally with a few extra examples to make it very clear.  Thanks for identifying it.

PR #611, opened 2021-01-18
Major nesting fix to file layout of jsguide.html

MattMarquise:
In this change I have fixed most of the nesting issues in an effort to simplify the layout of the jsguide.html file, all while promoting cleaner code which is crucial for an open source project such as this one.

### DIFF
--- a/jsguide.html
+++ b/jsguide.html
@@ -1,3611 +1,5308 @@
 <!DOCTYPE html>
 <html lang="en">
-<head>
-<meta charset="utf-8">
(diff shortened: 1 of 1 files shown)


MattMarquise:
@googlebot I signed it!

shicks:
Thank you for the contribution.  Unfortunately, we cannot accept it.  These style guides are generated from markdown and the HTML is not intended to be consumed by humans.  While we can _sometimes_ consider mirroring content changes/corrections internally, it's simply not feasible to adjust this generated output.

PR #612, opened 2021-01-18
Simple fix for nesting of tags.

MattMarquise:
This change is simply to fix an issue with improper tag nesting.


shicks:
These HTML are auto-generated, so it's not feasible to change the indentation.

PR #615, opened 2021-01-23
Add missing line in "Names and Order of Includes"

lengly:
Below line said: "Separate each non-empty group with one blank line."
So I think to avoid misunderstanding, we should add this missing line


shicks:
I've requested this change internally.  Thanks for pointing this out.  It should hopefully get updated in the next push.

PR #618, opened 2021-01-24
Another simple nesting fix.

MattMarquise:
Can we please begin working to make each file's code cleaner using basic nesting rules?

### DIFF
--- a/google-r-style.html
+++ b/google-r-style.html
@@ -1,17 +1,17 @@
 <!DOCTYPE html>
 <html>
-<head>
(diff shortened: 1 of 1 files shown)


shicks:
Thank you for the contribution.  Unfortunately, we cannot accept it.  These style guides are generated from markdown and the HTML is not intended to be consumed by humans.  While we can _sometimes_ consider mirroring content changes/corrections internally, it's simply not feasible to adjust this generated output.

PR #622, opened 2021-02-02
Back to top button

JackZeled0n:
Add back to top into htmlcss guide file


JackZeled0n:
@googlebot I signed it!

tonyruscoe:
Thanks for the suggestion, but I’m unfortunately unable to accept this pull request because the markup is automatically generated from a shared template we maintain internally.

JackZeled0n:
Is there any way to help you with maintenance?

tonyruscoe:
Not with the markup used by the style guides, but thanks for showing an interest!

PR #624, opened 2021-02-21
Do not recommend the `function` keyword. Do not require a "main"

IsaacG:
... at least externally

### DIFF
--- a/shellguide.md
+++ b/shellguide.md
@@ -1165,8 +1165,10 @@ may be done before declaring functions.
 
 ### main
 
-A function called `main` is recommended for scripts long enough
-to contain at least one other function.
+A dedicated entry point is required for scripts long enough to contain
+at least one other function. The name
+[`main`](https://en.wikipedia.org/wiki/Entry_point) is commonly used in
+other languages, so it will be familiar to most programmers.
 
 In order to easily find the start of the program, put the main program
 in a function called `main` as the bottom most function.
--- a/shellguide.md
+++ b/shellguide.md
@@ -196,7 +196,7 @@ Example:
 # Arguments:
 #   None
 #######################################
-function cleanup() {
+cleanup() {
   …
 }
 
@@ -209,7 +209,7 @@ function cleanup() {
 # Outputs:
 #   Writes location to stdout
 #######################################
-function get_dir() {
+get_dir() {
   echo "${SOMEDIR}"
 }
 
@@ -220,7 +220,7 @@ function get_dir() {
 # Returns:
 #   0 if thing was deleted, non-zero on error.
 #######################################
-function del_thing() {
+del_thing() {
   rm "$1"
 }
 ```
@@ -1006,8 +1006,8 @@ echo $(( hr * 3600 + min * 60 + sec )) # prints 7530 as expected
 
 Lower-case, with underscores to separate words. Separate libraries with
 `::`. Parentheses are required after the function name. The
-keyword `function` is optional, but must be used consistently
-throughout a project.
+keyword `function` is optional but not POSIX. Consistency throughout a project
+trumps POSIX.
 
 If you're writing single functions, use lowercase and separate words
 with underscore. If you're writing a package, separate package names
@@ -1028,8 +1028,7 @@ mypackage::my_func() {
 ```
 
 The `function` keyword is extraneous when "()" is present
-after the function name, but enhances quick identification of
-functions.
+after the function name and is not POSIX compliant.
 
 <a id="s7.2-variable-names"></a>
 
@@ -1166,7 +1165,7 @@ may be done before declaring functions.
 
 ### main
 
-A function called `main` is required for scripts long enough
+A function called `main` is recommended for scripts long enough
 to contain at least one other function.
 
 In order to easily find the start of the program, put the main program


IsaacG:
@googlebot I signed it!

shellguide.md
```diff
@@ -1006,8 +1006,8 @@ echo $(( hr * 3600 + min * 60 + sec )) # prints 7530 as expected
 
 Lower-case, with underscores to separate words. Separate libraries with
 `::`. Parentheses are required after the function name. The
-keyword `function` is optional, but must be used consistently
-throughout a project.
+keyword `function` is optional but not POSIX. Consistency throughout a project
```
dimo414, line 1009: IMO I'd rather this be a footnote, maybe something like: ``[^posix]: Note that POSIX requires the `function` keyword; if your project requires POSIX-compatibility the `function` keyword will be needed.`` I feel it detracts from the point to call it out inline (since anyone requiring POSIX-compatibility doesn't need this advice to know that the keyword is necessary for their script to parse correctly).

dimo414 (commented):
Thanks, if you don't mind I'll replicate these changes internally and then push it back out.

shellguide.md
vapier: i think you have this backwards: `function` is *not* in POSIX, so a POSIX-compliant script would never use the keyword.

i think keeping this section as-is is OK (since it only talks about consistency), and leaving the POSIX bit to the section below.

shellguide.md
```diff
@@ -1166,7 +1165,7 @@ may be done before declaring functions.
 
 ### main
 
-A function called `main` is required for scripts long enough
+A function called `main` is recommended for scripts long enough
```
vapier: i disagree with this: if your script is long enough to add functions, you need to be using `main`.  that's a good indicator that the script is growing too big and is going to get into the hard-to-untangle mess of not mixing code in global scope with functions and definitions.

shellguide.md
```diff
@@ -196,7 +196,7 @@ Example:
 # Arguments:
 #   None
 #######################################
-function cleanup() {
+cleanup() {
```
vapier, line 199: i've posted a CL internally to strip the `function` keywords from these few examples

shellguide.md
eli-schwartz: Not sure what you're talking about here.

If you mean, people might violate the rule elsewhere in this style guide:
> Put all functions together in the file just below constants. Don’t hide executable code between functions. Doing so makes the code difficult to follow and results in nasty surprises when debugging.

Then you're completely wrong, because you're suggesting that part of the style guide (use of `main`) only exists in order to work around people not following the rest of the style guide.

If you mean that code inside `main() { ...; }` avoids having code in global scope getting mixed up with and erroneously accessed by other functions, then you're still completely wrong... shell scoping rules are that code in a `main` function is global from the perspective of functions invoked by it... only once you leave `main "$@"` and proceed to EOF does it go out of scope.

There is zero benefit to `main()` functions in bash other than the stylistic "looks familiar to C programmers" point. I would argue that it should be warned against, not recommended or required, but lowering it to an opinionated Google internal-use-only *recommendation* is a fair compromise if it means innocent users on the internet stop thinking they have to do this because google told them to.

shellguide.md
dimo414: Yes, I misread the added text to mean "required" not "unusable" (so there's some wordsmithing to do here regardless). I'm OK with leaving the text as-is.

shellguide.md
vapier: no one is forcing you to use this style guide.  if you don't like it, move on.

forcing all the business logic into functions helps drive people to writing isolated code blocks with explicit arguments rather than intermingling code & defs, and relying on ambient state for arguments.

the idea that "main funcs are for C programmers" makes no sense.  you are aware that the main func/concept exists in plenty of languages beyond C/C++ right ?

eli-schwartz: > no one is forcing you to use this style guide. if you don't like it, move on.

WAIT. Are you recommending that if someone disagrees with one point on the style guide, they should NOT argue that the point is wrong, but instead, they should "stop using the style guide and move on"?

> the idea that "main funcs are for C programmers" makes no sense. you are aware that the main func/concept exists in plenty of languages beyond C/C++ right ?

It does, but not in shell. Do I need to enumerate the entire list of languages bash doesn't need to emulate here?

> forcing all the business logic into functions helps drive people to writing isolated code blocks with explicit arguments rather than intermingling code & defs, and relying on ambient state for arguments.

And once again, if your code already obeys the guidance
> Put all functions together in the file just below constants. Don’t hide executable code between functions. Doing so makes the code difficult to follow and results in nasty surprises when debugging.

then you will not be intermingling code & defs via the explicit rule "don't intermingle code & defs", no need to introduce another rule "do X because if you do, then as a side effect you won't intermingle code & defs"

"isolated code blocks with explicit arguments" is not going to happen either way, if someone is writing code like this:

```
file=/path/to/bar.txt

step1() {
    flob --thing-one="$file"
}
step2() {
    zindle "$file"
}

step1
step2
```
Rewriting it to 

```
file=/path/to/bar.txt

step1() {
    flob --thing-one="$file"
}
step2() {
    zindle "$file"
}

main() {
    step1
    step2
}

main "$@"
```
is not going to help anything here. And that's currently all the style guide tells people to do.

However, including an actual style guide recommendation "functions should never rely on global state, but accept arguments for that instead", would actually help. Then people would be encouraged to write:

```
step1() {
    flob --thing-one="$1"
}
step2() {
    zindle "$1"
}

file=$1; shift

step1 "$file"
step2 "$file"
```

No need for `main()` here, at least no just for this goal.

If your goal is to convince people "encapsulate functions and make them communicate via arguments instead of state", then you're currently dismally failing, because you don't mention this principle at all.

eli-schwartz: Regardless, I would like to reiterate that

> as well as allowing you to define more variables as local (which can’t be done if the main code is not a function

is utterly wrong. Wrapping code in main() does not permit you to define more variables as local. You can define them as local all you want, but they are localized to `main()` and functions invoked by `main()`, so they are NOT local in practice if the entire program runtime is running through `main()` as those variables never go out of scope.

IsaacG: My goal with this change was more about the function _name_ than the idea of having an entry point function.

I suppose this comes down to how much the shell style guide is about providing best practices to the general community vs making arbitrary style requirements. The section on functions makes a soft recommendation on sticking to POSIX-compliant function definitions. The _main_ section makes a pretty hard requirement on using a C-style function name. I completely agree with the value of having an entry point function as an entry point. _Requiring_ that it be called "main" seems pretty strongly opinionated about a name that little meaning.

Reverting to "no one is forcing you to use this guide" doesn't further the goal of creating a guide that benefits the guide and can be used by the broader public to help raise the quality of code being written out there.

vapier: @eli-schwartz you need to take it down a peg or two.  if you want to get into fights, this isn't the project for you.  i'm not wasting more time if you're just going to insist on being combative.  it is with that framing that i recommend you find another styleguide if this one doesn't meant your requirements.

@IsaacG i don't think one can really write a styleguide that is entirely "best practices" -- by its nature, it's going to get into opinionated spaces.  i'm not particularly wedded to the name `main`, but it isn't C-specific at all.  [it shows up in many other languages too](https://en.wikipedia.org/wiki/Entry_point) which makes it a convention that people are likely to be familiar with.  i'm not aware of any other common entry-point that crosses multiple language boundaries.

since we agree an entry point has to be declared, and `main` is the one people are most likely to be familiar with compared to any other name, i'm not seeing the problem with using it.  do you have a situation where you really really want it to be using a different name ?

IsaacG: If I'm writing a script to, say, process a file, I may want to name my entry point `process`. Suggesting `main` makes sense but *requiring* the entry point be named `main` seems pretty heavy handed. What's the benefit of requiring it be `main` vs recommending it?

It's not my guide, though, so I'll leave it to the maintainers to pass the decision on this one. I'm just making a suggestion here.

vapier: > What's the benefit of requiring it be `main` vs recommending it?

you already know the answer to that :).  anytime people read "recommended" they see "i don't have to follow this rule".  the people we're targeting don't operate on subtly and find any excuse to deviate (remember the guy who wanted the guide written using legalese?).

if you wanted to wordsmith it to make it sound softer while technically not, i wouldn't say no.  maybe something like:
> A dedicated entry point is required for scripts long enough to contain at least one other function. The name [`main`](https://en.wikipedia.org/wiki/Entry_point) is commonly used in other languages, so it will be familiar to most programmers.

IsaacG: Updated using that blurb ;)

shicks:
@vapier this seems to have met some resistance internally.  Should we push that CL forward or close this as infeasible?

IsaacG:
@vapier  Ping

PR #629, opened 2021-03-22
Java: Recommend using inline `{@return }` for Java >= 16

Marcono1234:
[JDK-8075778](https://bugs.openjdk.java.net/browse/JDK-8075778) (note that the issue description does not match the final implementation) introduced an [inline `{@return }` tag](https://docs.oracle.com/en/java/javase/16/docs/specs/javadoc/doc-comment-spec.html#return) for Java 16. This tag should be preferred to avoid repetition, e.g.:
```java
/**
 * Returns the name.
 * @return the name
 */
```
Instead:
```java
/**
 * {@return the name}
 */
```

The current recommendation of the style guide to omit the `@return` tag would yield a javadoc warning and would not create a description for the return type.

### DIFF
--- a/javaguide.html
+++ b/javaguide.html
@@ -83,7 +83,7 @@ anywhere in a source file. This implies that:</p>
 <h4 id="s2.3.2-special-escape-sequences">2.3.2 Special escape sequences</h4>
 
 <p>For any character that has a
-<a href="http://docs.oracle.com/javase/tutorial/java/data/characters.html">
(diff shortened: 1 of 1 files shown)


shicks:
Any style guide changes need to be compatible with Google's internal development story.  Unfortunately, Google is a few years behind on Java language versions, so Java 16 is still a long way off, and we cannot reasonably accept a style guide change that would require using an unusable version.  Additionally, we generally have problems accepting any substantive changes from externally because style guide changes must be approved internally and can only then be pushed out.

PR #655, opened 2021-11-29
Missing some titles in the jsguide.html file, under "3 Source file structure" title + seeing the table of contents with delay .

ellietms:
Hi, I realised we are missing some titles under the 3 Source file structure section.
[JavascriptGuide](https://google.github.io/styleguide/jsguide.html)

it shows us 3.3 and we do not have 3.3.1, 3.3.2 but we have 3.3.3.
The Table contents are being created inside of the "include/jsguide.js" with pure javascript.
I added some comments that where I think the issue is happening.

one more thing is when I refresh the page, it takes longer for me to see all the links in my table of contents.
It is slow.

### DIFF
--- a/include/styleguide.js
+++ b/include/styleguide.js
@@ -1,10 +1,9 @@
 TocTypeEnum = {
   VERTICAL: 1,
-  HORIZONTAL: 2
+  HORIZONTAL: 2,
 };
 
 function CreateTOC(tocElement) {
-
   // Find the toc element DIV. We'll place our TOC there.
   var toc = document.getElementById(tocElement);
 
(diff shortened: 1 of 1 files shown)


devvsakib (approved):
it's not mandatory to use a comma after the last element, also single quote or double quote doesn't matter. 
LGTM

vapier (changes requested):
it's hard to tell functional from style changes in here.  please create a separate PR/commit for just the style changes.

venu-dhannoju:
It looks like the last review requested splitting the functional changes from the style-only changes to make review easier. Since there hasn't been activity for a while, is this PR still being worked on?

If so, updating the branch and separating the formatting changes into a dedicated commit (or PR) may help move the review forward.

ellietms:
> It looks like the last review requested splitting the functional changes from the style-only changes to make review easier. Since there hasn't been activity for a while, is this PR still being worked on?
> 
> If so, updating the branch and separating the formatting changes into a dedicated commit (or PR) may help move the review forward.

Hey, I do not think so

vapier:
if you don't intend on working on it, let's close it out

PR #680, opened 2022-03-22
Typo fix: change 7.3.4 to 7.3.3 in javaguide.html

Arielwz:
1. There is no 7.3.3, so 7.3.4 is changed to 7.3.3. 
2. Delete useless hyperlink <a> for previous 7.3.3.

only then copied here.

through the usual process.

### DIFF
--- a/javaguide.html
+++ b/javaguide.html
@@ -1189,7 +1189,6 @@ punctuated as if it were a complete sentence.</p>
 incorrect, and should be changed to
 <code class="prettyprint lang-java">/** Returns the customer ID. */</code>.</p>
 
-<a name="s7.3.3-javadoc-optional"></a> 
 <h3 id="s7.3-javadoc-where-required">7.3 Where Javadoc is used</h3>
 
 <p>At the <em>minimum</em>, Javadoc is present for every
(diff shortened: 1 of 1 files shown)


dimo414:
Thanks for the heads-up. There is actually a 7.3.3, it's just not published externally (which is certainly confusing). I'll try to re-number the sections so it's less odd-seeming, though it may be easier to just leave the numbering be so as to not break existing links.

dimo414:
Thank you again for flagging this, but for the time being we're just going to leave this be. Silly as it may sound, it's a bit tedious to fix this both internally and externally, and we've decided it's just not worth the effort at the moment. If/when we do a larger refresh we'll try to account for skipped subsections more carefully.

javaguide.html
```diff
@@ -1189,7 +1189,6 @@ <h3 id="s7.2-summary-fragment">7.2 The summary fragment</h3>
 incorrect, and should be changed to
 <code class="prettyprint lang-java">/** Returns the customer ID. */</code>.</p>
 
-<a name="s7.3.3-javadoc-optional"></a>
```
dimo414, line 1192: Note this is not a link but an anchor, and shouldn't be removed because existing links may still use it.

Arielwz:
Thanks for letting me know.

PR #681, opened 2022-03-31
Update pyguide.md

akashhansda:
Fix : invalid links

### DIFF
--- a/pyguide.md
+++ b/pyguide.md
@@ -544,7 +544,7 @@ variables defined in enclosing scopes.
 
 Allows definition of utility classes and functions that are only used inside of
 a very limited scope. Very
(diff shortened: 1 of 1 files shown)


ProKashif (approved):

kbakdev (approved):

gpshead:
these weren't *invalid* links, but were needlessly redirecting through the search engine.

historical explanation: we used to do this for links from internal resources as an attempted general policy a decade+ ago to avoid the HTTP `Referer:` headers "leaking" the URLs of internal-only documents to the external public web sites.  Not really sensible in the modern age and never done completely.  A few things like thing lingered.  It turns out we cleaned these out of pyguide.md internally within the last few months anyways.  I just pushed an update so the're gone now.  no more need for this PR. :)

thanks!

PR #685, opened 2022-04-18
Update shellguide.md

ss-o:
Fix incorrect statement.

Additionally, the content seems to have mixed/incorrect terms as it would have the same meaning.
As a style guide is correct, but the context should be updated/corrected. 

1. bash
2. shell

If someone agrees, I would like to help make corrections as  I would like to include this
to my projects as a guide.

---

only then copied here.

through the usual process.

### DIFF
--- a/shellguide.md
+++ b/shellguide.md
@@ -39,7 +39,7 @@ Section                                                                         
 
 ### Which Shell to Use
 
-Bash is the only shell scripting language permitted for
+Bash is shell scripting language permitted for
 executables.
 
 Executables must start with `#!/bin/bash` and a minimum


dimo414:
> Bash is shell scripting language permitted for executables.

Perhaps you meant "Bash is _a_ shell scripting language permitted for executables." but that is not correct. We intentionally only recommend Bash, as the following paragraph explains:

> Restricting all executable shell scripts to `bash` gives us a consistent shell language that’s installed on all our machines.

If you don't work at Google you are obviously not bound by our Style Guide, but within Google this is the intended policy.

PR #708, opened 2022-08-26
Fix typo

Bakr-Ali:
**Fixed the following typos:**
don't it?  ->  doesn't it?
makes layout  ->  makes the layout


ProKashif (approved):

kbakdev (approved):

devvsakib (approved):

vapier:
thanks for checking

the first typo has been fixed in commit 1ae7350b45445b3a3a289c5fe756ab7c80014fe5

the second one i don't think is wrong ... i think both can be argued as "correct".  since that's how it's written internally, i'll leave it as-is externally.

PR #716, opened 2022-10-18
cppguide: fix html markup & namespace

parmaski:
- Added missing `std::` namespace to classes in the standard library e.g., `string_view`.
- Added missing &lt;code&gt; tag to keywords and file extensions where missing.
- Fixed all paragraphs to use &lt;p&gt; tag.
- Removed unnecessary &lt;div&gt; tags.

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -174,9 +174,8 @@ input.</p>
 <p>Do not use
   <a href="#Nonstandard_Extensions">non-standard extensions</a>.</p>
 
-  <div>Consider portability to other environments
-before using features from C++14 and C++17 in your project.
-</div>
+<p>Consider portability to other environments
+before using features from C++14 and C++17 in your project.</p>
 
 <h2 id="Header_Files">Header Files</h2>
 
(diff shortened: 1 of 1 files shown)


tituswinters (commented):
Thanks for the PR! The source of truth for this is internal, but I'm porting most of these changes over to that doc so these will be reflected on the next update. 

That said, there's a couple things that will probably be a little iffy:
* We're not explicitly saying `std::string_view` yet because of minor compatibility differences between `absl::string_view` and `std::string_view` (mostly regarding construction from a single nullptr). So we're leaving the unqualified names in place as an intentional ambiguity.
* There's a bunch of random <div>s that are coming from the upstream source-of-truth having divs for internal/external-only stanzas. Unless those are causing harm, I think they're relatively likely to keep popping up. 

I *think* I've tidied up the full qualification for other types and other general use of <code> and <p> tags.

parmaski:
Thank you for merging to the source-of-truth! It's helpful for machine translation.

tituswinters:
It did get committed up-stream, I expect we'll sync this in the next month (but could do so sooner if it matters to you).

Do the stray <div> tags cause problems?

parmaski:
Thanks a lot. I just wait the update.

&lt;p&gt; and &lt;div&gt; tags are not big problem.
I didn't know whether the HTML was hand-coded or auto-generated.
In the latter case, I don't mind.

kbakdev (approved):

devvsakib (approved):

PR #717, opened 2022-10-20
Update javaguide.css

ibrahimmemonn:


devvsakib (approved):

vapier (changes requested):
please describe what it is you're trying to do/fix here.  there is no "array" html element afaict.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/array -> 404

nor are we defining a custom "array" element.

PR #718, opened 2022-10-21
fix: broken link due to spacing

jukrb0x:
the first paragraph, 
> "Say what you mean, simply and directly." - [Brian Kernighan] (https://en.wikipedia.org/wiki/The_Elements_of_Programming_Style)

which brokes the link since the markdown syntax is incorrect, I think the best practice for this best practice is to fix the best practice.

this PR fixes a broken link since the space between `[]` and `()` in the markdown, should not appear here.

### DIFF
--- a/docguide/best_practices.md
+++ b/docguide/best_practices.md
@@ -1,7 +1,6 @@
 # Documentation Best Practices
 
-"Say what you mean, simply and directly." - [Brian Kernighan]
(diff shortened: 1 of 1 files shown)


vapier:
sorry, missed your earlier PR.  this has been fixed in commit 7b9a1c1330973e14f9486af1d2760f798eff1ce9 now.

PR #727, opened 2022-11-09
Update ts styleguide with latest google version

Cabbageboy:
Updates from latest internal Typescript StyleGuide


mprobst (approved):
Approving as a TS/JS tech lead.

PR #738, opened 2022-12-06
Small clarification to class docstring examples

anthonyburdi:
Clarify that the first code snippet is the example to follow, and the second is not. This is clear from context if you read the paragraph above but not if you only look at the examples.

only then copied here.

through the usual process.


gpshead:
thanks.  the yes/no markers had been unintentionally omitted from those.  an internal change addressing this was included as part of https://github.com/google/styleguide/commit/ec302b363edada1d696c476dcaf69f310755f823.

PR #750, opened 2023-01-09
Error print from https://github.com/cpplint/cpplint fixed

haobo724:
something wrong with function `sys.stderr.write()`  in file cpplint.py
when I use cpplint.py from git@github.com:google/styleguide.git 
the console always shows an Error msg:
`TypeError('write() argument must be str, not bytes')`
But if i use the cpplint.py from https://github.com/cpplint/cpplint  everything goes well.

By comparison I found that it was a matter of how some of the error messages were called by function.
So I simply copied the file from https://github.com/cpplint/cpplint and rewrote the old one

### DIFF
--- a/cpplint/cpplint.py
+++ b/cpplint/cpplint.py
@@ -41,30 +41,55 @@ We do a small hack, which is to ignore //'s with "'s after them on the
 same line, but it is far from perfect (in either direction).
 """
 
+# cpplint predates fstrings
+# pylint: disable=consider-using-f-string
+
+# pylint: disable=invalid-name
+
 import codecs
 import copy
 import getopt
+import glob
+import itertools
 import math  # for log
 import os
 import re
 import sre_compile
 import string
 import sys
-import unicodedata
 import sysconfig
+import unicodedata
+import xml.etree.ElementTree
+
+# if empty, use defaults
+_valid_extensions = set([])
+
+__VERSION__ = '1.6.1'
 
 try:
+  #  -- pylint: disable=used-before-assignment
   xrange          # Python 2
 except NameError:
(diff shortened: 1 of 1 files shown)


devvsakib (approved):

tkruse:
As the maintainer of [cpplint on pypi](https://github.com/cpplint/cpplint), I would recommend NOT merging this PR.

As far as I am aware, Google still has an internal version of cpplint.py which has changes not in this github repository, and merging this PR would make it more difficult to later publish more internal changes (even if it seems increasingly unlikely that this would ever happen). See https://github.com/google/styleguide/pull/528#issuecomment-592315430

As much as I'd like to see the work I have done in https://github.com/cpplint/cpplint be used widely, I'd very much rather keep the hope alive that someday somebody at Google will update the github version of cpplint.py with what Google uses internally.

If Google wanted to abandon the public cpplint.py, it should just be removed from this repo to send people to use https://github.com/cpplint/cpplint, rather than copying back changes from there to here.

Even if for some reason this github repo should incorporate the changes from https://github.com/cpplint/cpplint , then it would be much more reasonable to do so while maintaining the history (it's tricky but possible), so that for each change there is still a separate commit with the context.

Else merging this should have a different commit message than "[Error print from](https://github.com/google/styleguide/pull/750/commits/18203a1009223a3ea9f3b55c7d89710166d34439) https://github.com/cpplint/cpplint [fixed](https://github.com/google/styleguide/pull/750/commits/18203a1009223a3ea9f3b55c7d89710166d34439)", it should rather be the summary of changes form the changelog: https://github.com/cpplint/cpplint/blob/develop/changelog.rst

vapier:

PR #778, opened 2023-06-01
The base_name should fully match the exclude_files pattern

kenkangxgwe:
Previously, a path in the same directory as the CPPLINT.cfg is excluded if one of its prefixes matches the pattern, which leads some confusing cases like "exclude_files=l" excluding "libs". With this change we should change the pattern to "exclude_files=l.*" to achieve the same result.


vapier:

PR #787, opened 2023-08-22
Update C++ style guide.

zetafunction:
- Update caveats for `thread_local` usage, particularly around the risk of destruction order issues.
- Provide explicit guidance for situations where `bit_cast` may be a better fit than `reinterpret_cast`.
- Emphasize that kConstant-style naming can still be used for `const` automatic variables that are initialized at runtime, but only if the resulting variable has the same value with each evaluation (i.e. it does not depend on runtime inputs).
- Clarify what sorts of details belong in file-level comments vs comments for individual abstractions.
- Update TODO examples to reflect the preferred styling, from most preferred to least preferred.

only then copied here.

through the usual process.

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -1044,9 +1044,12 @@ thread, rather than once at program startup. This means that
 other <code>thread_local</code> variables are subject to the same
 initialization-order issues as static variables (and more besides).</p>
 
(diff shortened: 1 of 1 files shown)


tituswinters:
Thanks!

PR #788, opened 2023-09-02
Updated README.md for easy undersanding

Tomilola-ng:
only then copied here.

through the usual process.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,52 +1,53 @@
-# Google Style Guides
+# Google Style Guides 📝
 
 Every major open-source project has its own style guide: a set of conventions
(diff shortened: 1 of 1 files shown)


vapier:
seems like the enhancement here is adding a lot of emojis ?  some of the language ones seem obvious (e.g. Python), but others not so much (e.g. lisp & C#).  overall, i'm not sure this adds much value, and in some cases, seems excessive (e.g. use of 👇 just to point at the next line of text which is where readers are already going to be looking).  thanks for the suggestions, but i think we'll keep things as-is.

PR #789, opened 2023-09-13
Go: update interface{} to any

### LINKED ISSUE #770: Go: update `interface{}` to `any`
After go.1.18 you can use `any` and `interface{}` are interchangeably.

I would love the change so when I do a ctrl+F and look for styles around how `interface`s should work I don't get false positives.

The style guide [already prefers the use](https://google.github.io/styleguide/go/decisions#use-any).

I would just want the examples here updated
* https://google.github.io/styleguide/go/decisions#assertion-libraries
* https://google.github.io/styleguide/go/best-practices#parameters-and-configuration

NextThread:
Fixes: #770

### DIFF
--- a/go/best-practices.md
+++ b/go/best-practices.md
@@ -1255,7 +1255,7 @@ information to the reader:
 // string.
 //
 // format is the format, and data is the interpolation data.
-func Sprintf(format string, data ...interface{}) string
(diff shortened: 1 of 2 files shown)


gaal:
Thanks, this repo is not accepting pull requests. We'll update this upstream and push it when we do a periodic update.

PR #793, opened 2023-09-28
Refactored file jsguide.js

khaidao000:
Refactoring includes:

Using querySelector for simplicity where possible.
Utilizing Array.prototype.forEach.call to directly iterate over NodeList.
Utilizing const for variable declarations where possible.
Simplifying if conditions.
Grouping similar logic and segregating different logic with line breaks for readability.

### DIFF
--- a/include/jsguide.js
+++ b/include/jsguide.js
@@ -1,63 +1,45 @@
+// Defining a style guide initialization function
 window.initStyleGuide = function(init) {
-  // Runs the callback on every element matched by the query selector.
+  // Utility function to iterate over elements matching a query selector
   function find(querySelector, callback) {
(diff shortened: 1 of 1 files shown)


vapier:
thanks for the suggestions.  you'll have to complete the CLA first though please.

vapier (changes requested):
please break this up into multiple changes.  there's good stuff in here, but squashing all of this rewrite into a single commit makes it difficult/impossible to reasonably review, and makes it hard to detect bugs or unintentional changes in behavior.

vapier:
feel free to re-open or open a new PR if you plan on addressing the feedback

PR #794, opened 2023-10-11
Update ts styleguide with latest google version

freshp86:


freshp86:
@shicks Could you help assign a reviewer to this PR? I've tried a couple channels internally without success so far. Asking since I see that you had merged the original version of this file at https://github.com/google/styleguide/pull/610. Thank you!

PR #795, opened 2023-10-12
Update C++ style guide for C++20.

zetafunction:
- Encourage single line nested namespace declarations.
- Reference and allow `constinit` in the relevant sections
- Update operator overloading guidance for comparison operators: prefer only to overload `==` and optionally `<=>` when there is an obvious ordering, and allow the compiler to derive the other comparison operators.
- Discourage prefixing `uint32_t`, et cetera with `std::`.
- Document when and how to use concepts:
  - Use `requires` expressions rather than the alternatives, e.g. a template parameter.
  - Do not reimplement existing concepts/traits.
  - Do not expose concepts across API boundaries.
  - Do not use concepts unnecessarily.
  - Do not implement concepts that are not compile-time checkable.

I tried to sum up the concepts changes in the commit description without going into too much detail, since I suspect that's what people will be most interested in.

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -21,8 +21,8 @@ this power brings with it complexity, which in turn can make
 code more bug-prone and harder to read and maintain.</p>
 
 <p>The goal of this guide is to manage this complexity by
-describing in detail the dos and don'ts of writing C++ code
-. These rules exist to
+describing in detail the dos and don'ts of writing C++
+code. These rules exist to
 keep  the code base manageable while still allowing
 coders to use C++ language features productively.</p>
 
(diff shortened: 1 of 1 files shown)


cppguide.html
```diff
@@ -1466,9 +1466,9 @@ <h3 id="Structs_vs._Classes">Structs vs. Classes</h3>
 be present; however, these methods must not require or enforce any
 invariants.</p>
 
-<p>If more functionality or invariants are required, a
-<code>class</code> is more appropriate. If in doubt, make
-it a <code>class</code>.</p>
+<p>If more functionality or invariants are required, or struct has wide visibility and expected to
```
pkasting, line 1469: This grammar seems problematic (maybe `...or the struct has wide visibility and is expected...`?)

zetafunction: I agree, but that should be fixed with an internal update and then syncing the public style guide to it. If you get around to it before me, let me know and I can update the PR.

pkasting: I'm not going to get to this this week.

tituswinters (approved):

PR #802, opened 2024-02-01
Correction: jsguide.html - Changed null to undefined

KumarSidharth:
## This change is "Not important".
Please don't review this leaving your work behind.

## Change description 
Keys not defined in an Object are undefined.
``` javascript
const o = {width: 42};
o.maxWidth === null // false
o.maxWidth === undefined // true
```
image


vapier:
the example is using `!=`, not `!==`, so there is coercion involved, so I think this is written correctly following the guidance:

https://google.github.io/styleguide/jsguide.html#features-equality-checks-exceptions

PR #804, opened 2024-02-12
Update C++ style guide.

zetafunction:
- Give more explicit guidance about when angle bracket includes should be used.
- Expand the guidance for disallowing const reference parameters that outlive the call to *all* references, const or mutable; instead, these parameters should be pointers.
- Add a brief section about how concepts should be named

There are also additional minor formatting changes or updating recommendations to prefer std over absl.

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -414,6 +414,18 @@ should be included as:</p>
 <pre>#include "base/logging.h"
 </pre>
 
+<p>Headers should only be included using an angle-bracketed path if the library
+requires you to do so. In particular, the following headers require angle
+brackets:</p>
+
+<ul>
+<li>C and C++ standard library headers (e.g. <code>&lt;stdlib.h&gt;</code>
+  and <code>&lt;string&gt;</code>).</li>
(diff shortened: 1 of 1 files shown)


cppguide.html
```diff
@@ -1093,13 +1103,11 @@ <h3 id="thread_local">thread_local Variables</h3>
...
-
-
-  <a href="https://github.com/abseil/abseil-cpp/blob/master/absl/base/attributes.h">
-  <code>ABSL_CONST_INIT</code></a>
+  <a href="https://en.cppreference.com/w/cpp/language/constinit">
+    <code>constinit</code></a>
   (or <code>constexpr</code>, but that should be rare):</p>
 
-  <pre>  ABSL_CONST_INIT thread_local Foo foo = ...;
+  <pre>   constinit thread_local Foo foo = ...;
```
pkasting, line 1110: Extra space?

pkasting (commented):
(Maybe the below is an upstream bug)

cppguide.html
zetafunction: Yes, this is an upstream bug; the external one is programmatically extracted from the internal one.

geoffromer (approved):

PR #806, opened 2024-03-03
C++: Added ordered list under the heading Goals of the Style Guide and add …

Mustafa-Hassan2001:
…content about scoping under it's heading in cppguide.html file

These style guides are copies of Google's internal style guides to
assist developers working on Google owned and originated open source
projects. Changes should be made to the internal style guide first and
only then copied here.

Unsolicited pull requests will not be merged and are usually closed
without comment. If a PR points out a simple mistake — a typo, a broken
link, etc. — then the correction can be made internally and copied here
through the usual process.

Substantive changes to the style rules and suggested new rules should
not be submitted as a PR in this repository. Material changes must be
proposed, discussed, and approved on the internal forums first.

PR #821, opened 2024-04-13
Content and list view enhancement in cppguide.html

MohsinMalik2:
Fixes #806 

Changes

1. Added ordered list for major headings of Goals of the Style Guide

2. Content added for the heading "Scoping".

- Scope in programming refers to the visibility and accessibility of variables, functions, and other identifiers within a program. It plays a crucial role in preventing naming conflicts by limiting the visibility of identifiers to specific regions of code, such as within a function, block, or namespace.

- Global scope refers to identifiers accessible throughout the entire program, while local scope refers to identifiers accessible only within a specific block or function. Nested scopes occur when one scope is contained within another, allowing identifiers declared in an outer scope to be accessed by inner scopes, but not vice versa.

- Understanding scoping rules is essential for writing maintainable and error-free code, as it helps prevent unintended side effects and promotes modular design.

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -57,7 +57,8 @@ necessary to change a rule in the guide.</p>
 
 <p>The goals of the style guide as we currently see them are as follows:</p>
 <dl>
-<dt>Style rules should pull their weight</dt>
+<ol>  
+<li><dt><b>Style rules should pull their weight:</b></dt></li>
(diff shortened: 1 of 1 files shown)


Fajar1901:
MohsinMalik2:content-enhancement-cppguide

https://github.com/google/styleguide/pull/821#issue-2241557364

PR #829, opened 2024-05-05
Remove extra lines that yield paragraph tags

dane:
I noticed inconsistent line spacing across the rendered bulleted list. These empty lines this PR removes were the cause.


gaal:
Thanks for the patch. This repo doesn't accept pull requests; we'd have to fix this internally. Unfortunately, though, this is probably an artifact of copybara stripping of internal references, so it may not be feasible.

dane:
No problem. I just wanted to put the mis-render on your radar.

PR #837, opened 2024-07-08
cpplint: declare bankruptcy

vapier:
We have not pushed updates to the public cpplint code from our internal version in years.  No one seems interested in picking it up either.

Since an open source community has formed & forked & updated it, let's delete our ancient version and just send people to that project now.


vapier:
@jdennett @fowles @dinord @gpshead @markmentovai

README.md
```diff
@@ -31,9 +31,11 @@ that apply to that project.
 *   [TypeScript Style Guide][ts]
 *   [Vim script Style Guide][vim]
 
-This project also contains [cpplint][cpplint], a tool to assist with style guide
-compliance, and [google-c-style.el][emacs], an Emacs settings file for Google
-style.
+This project also contains [google-c-style.el][emacs], an Emacs settings file
+for Google style.
+
+We used to host the [cpplint tool](https://github.com/cpplint/cpplint), but it
```
tkruse, line 37: This is slightly misleading. Google never hosted on that github org, it always hosted cpplint.py at https://github.com/google/styleguide/gh-pages/cpplint/cpplint.py, of which https://github.com/cpplint/cpplint has been forked

tkruse: This markup is slightly misleading, it could be read as if Google had hosted the fork at https://github.com/cpplint/cpplint. I would suggest:

```
We used to host the google internal cpplint tool, but won't be pushing updates.
Open source forks such as [cpplint on PyPi](https://github.com/cpplint/cpplint) exist.
```

vapier: good point.  i don't think mentioning PyPi makes sense, but i'll update the text.

PR #839, opened 2024-07-09
LICENSE: switch to CC-BY-3.0 #637

### LINKED ISSUE #637: License clarification
The `LICENSE` file in the project says it's an Apache license, but in the `README.md`, it says:
```
The style guides in this project are licensed under the CC-By 3.0 License, which encourages you to share these documents. See https://creativecommons.org/licenses/by/3.0/ for more details.
```
This seems to conflict. What does the Apache license cover?

vapier:
We've always said the style guides are licensed under CC-By 3.0 in our README file.  We use other licenses for code snippets, and list those in the respective file headers, but the guides are the dominant source in this repo, so list that in the LICENSE file.

Fixes #637.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,202 +1,319 @@
+Creative Commons Legal Code
 
-                                 Apache License
-                           Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+Attribution 3.0 Unported
 
-   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+    CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES NOT PROVIDE
+    LEGAL SERVICES. DISTRIBUTION OF THIS LICENSE DOES NOT CREATE AN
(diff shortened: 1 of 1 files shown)


vapier:
@dseomn @gpshead FYI

gpshead:
There's no need to ask me about anything in this repo anymore. go/epitaphs/gps
_(Fixing the license to be what makes sense for documentation makes sense to me but I have no say in the matter anymore)._

vapier:
sorry bud, missed that.  i'll stop spamming you :).

dseomn:
Personally, I'd really like some way to keep using the pylintrc in other projects (because that's the point of it, right?), but I have no stake in this decision as part of my job.

vapier:
nothing is stopping you from using the file elsewhere.  it's arguable whether one can copyright config files like this.  if one thinks so, based on the comments written, not the key=value parts, then the license would be that of pylint since that's where all the comments are coming from.  and pylint is GPL-2, not Apache-2.0.

vapier:
I updated our internal flow to add an Apache header to the pylintrc file and updated it here via #885

PR #888, opened 2025-02-26
go: Export the latest version of internal guide.

matttproud:
The updates encompass copy editing, enchanced guidance on package
sizing, fixing broken hyperlinks, simplifying Protocol Buffer message
guidance, clarifying context root behavior after (testing.TB).Context,
and disjoint table test.

### DIFF
--- a/go/decisions.md
+++ b/go/decisions.md
@@ -69,6 +69,9 @@ exceptions to this principle:
     reuse identifiers, as is done in [`syscall`]. This is expected to be very
     rare in most codebases.
 
+**Note:** Filenames of source code are not Go identifiers and do not have to
+follow these conventions. They may contain underscores.
+
 [`syscall`]: https://pkg.go.dev/syscall#pkg-constants
 
 <a id="package-names"></a>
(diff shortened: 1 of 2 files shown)


go/decisions.md
```diff
@@ -1065,7 +1068,7 @@ can get wrapped into an interface and thus become a non-nil value (see the
 func Bad() *os.PathError { /*...*/ }
 ```
 
-**Tip**: A function that takes a `context.Context` argument should usually
+**Tip**: A function that takes a [`context.Context`] argument should usually
```
chressie: Should be "**Tip:**".

chressie (approved):

go/decisions.md
matttproud: That should be fixed now.

chressie (approved):
LGTM

go/decisions.md
vapier: this should prob be handled in the copy.bara.sky file rather than modified by hand

chressie: I agree, but please note, that I also made the same change internally around the same time.

PR #914, opened 2025-04-07
Update C++ style guide

zetafunction:
- Expand guidance around usage of functions (and variables) with inline definitions.
- Change guidance around internal namespaces to reduce the likelihood of collisions.
- Expand naming guidance, with specific additional rules around:
  - nested namespaces not reusing well-known top-level names
  - template parameters
  - name aliases
- Clarify that constant-style naming also applies to static class constants.
- Other miscellaneous formatting and documentation improvements, e.g. additional Abseil TotW links.

only then copied here.

through the usual process.

### DIFF
--- a/cppguide.html
+++ b/cppguide.html
@@ -349,50 +349,94 @@ void test(D* x) { f(x); }  // Calls f(B*)
 <p>Try to avoid forward declarations of entities
 defined in another project.</p>
 
-<h3 id="Inline_Functions">Inline Functions</h3>
+<a id="Inline_Functions"></a>
+<h3 id="Defining_Functions_in_Header_Files">Defining Functions in Header Files</h3>
 
-<p>Define functions inline only when they are small, say, 10
-lines or fewer.</p>
+<p>Include the definition of a function at its point of declaration in a header file only
+when the definition is short. If the definition otherwise has a reason be in the header,
+put it in an internal part of the file. If necessary to make the definition ODR-safe, mark
+it with an <code>inline</code> specifier.</p>
 
 <p class="definition"></p>
-<p>You can declare functions in a way that allows the compiler to expand
-them inline rather than calling them through the usual
-function call mechanism.</p>
+<p>Functions defined in header files are sometimes referred to as "inline functions",
+which is a somewhat overloaded term that refers to several distinct but overlapping
+situations:</p>
+
+<ol>
+  <li>A <em>textually inline</em> symbol's definition is exposed to the reader at
+  the point of declaration.</li>
+  <li>A function or variable defined in a header file is <em>expandable inline</em>
+  since its definition is available for
+  <a href="https://en.wikipedia.org/wiki/Inline_expansion">inline expansion</a>
+  by the compiler, which can lead to more efficient object code.</li>
+  <li><em>ODR-safe</em> entities do not violate the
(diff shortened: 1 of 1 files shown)


zetafunction:
@geoffromer please take a look when you can; thanks!

cppguide.html
```diff
@@ -4245,29 +4296,57 @@ <h2 id="Naming">Naming</h2>
...
+while a local variable probably shouldn't explain what function it's within.</p>
+
+<p>Minimize the use of abbreviations that would likely be unknown to someone
+outside your project (especially acronyms and initialisms). Do not abbreviate by
+deleting letters within a word. When an abbreviation is used, prefer to
+capitalize it as a single "word", e.g., <code>StartRpc()</code> rather than
+<code>StartRPC()</code>.</p> As a rule of thumb, an abbreviation is probably OK
+if it's listed in
+
+Wikipedia. Note that certain universally-known abbreviations are OK, such as
```
geoffromer: It's confusing that this passage contains a `</p>` and a blank line, but they don't line up. I don't know if it's possible to avoid the blank line, but we should at least get rid of the `</p>`.

cppguide.html
```diff
@@ -4245,29 +4296,57 @@ <h2 id="Naming">Naming</h2>
...
+<code>i</code> for a loop index and <code>T</code> for a template
+parameter.
+
+<p>The names you see most frequently are not like most names; a small number of
+"vocabulary" names are reused so widely that they are always in context. These
+names tend to be short or even abbreviated and their full meaning comes from
+explicit long-form documentation rather than from just comments on their
+definition and the words within the names. E.g. <code>absl::Status</code> has a
+dedicated page, with short
+link <a href="https://goto.google.com/status">go/status</a>, documenting its
```
geoffromer: This is an internal link. It has a [public equivalent](https://abseil.io/docs/cpp/guides/status), but the fact that it's a short link seems like it might be load-bearing here, so I'm not exactly sure how to fix it. I'll confer with the style team.

geoffromer:
The problems I flagged should now be fixed upstream.

zetafunction:
Thanks, I believe all comments should be addressed now. I think I messed up the conversations though since I used a force push—sorry not really used to the GH contribution workflow.

geoffromer (approved):

cppguide.html
```diff
@@ -2171,7 +2222,7 @@ <h3 id="cpplint">cpplint</h3>
 how to run <code>cpplint.py</code> from their project
 tools. If the project you are contributing to does not,
 you can download
-<a href="https://raw.githubusercontent.com/cpplint/cpplint/HEAD/cpplint.py">
+<a href="https://raw.githubusercontent.com/google/styleguide/gh-pages/cpplint/cpplint.py">
```
aaronliu0130, line 2225: @zetafunction Was there a reason for this change? Since this PR was deployed, this has been linking to the [bankrupted](https://github.com/google/styleguide/pull/837) version... I'm guessing the bankruptcy PR was just never ported to the internal version?

zetafunction: This update is generated from the internal style guide. If someone makes a PR to update the external style guide but that change never makes it to the internal style guide, it's going to be overwritten by the next update.

vapier: i.e. I screwed up.  someone has posted a fix internally for it though, so it should make its way back out.

PR #916, opened 2025-04-22
docs: clarify map(parseInt) example with explicit radix

KeenoLee:
Use an arrow-function callback to explicitly forward only the string and a radix of 10 to parseInt:

// GOOD: we only forward the string (and an explicit radix)
['11','5','3'].map(n => parseInt(n, 10)); // [11, 5, 3]

### DIFF
--- a/tsguide.html
+++ b/tsguide.html
@@ -1561,8 +1561,8 @@ const numbers = ['11', '5', '10'].map(parseInt);
 <p>Instead, prefer passing an arrow-function that explicitly forwards parameters to
 the named callback.</p>
 
-<pre><code class="language-ts good">// GOOD: we only forward the string (and an explicit radix)
-const numbers = ['11', '5', '3'].map((n) =&gt; parseInt(n,10));
+<pre><code class="language-ts good">// GOOD: Arguments are explicitly passed to the callback
+const numbers = ['11', '5', '3'].map((n) =&gt; parseInt(n, 10));
 // &gt; [11, 5, 3]
 
 // GOOD: Function is locally defined and is designed to be used as a callback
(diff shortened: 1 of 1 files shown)


tsguide.html
```diff
@@ -1561,8 +1561,8 @@ <h4 id="functions-as-callbacks" class="numbered">Prefer passing arrow functions
 <p>Instead, prefer passing an arrow-function that explicitly forwards parameters to
 the named callback.</p>
 
-<pre><code class="language-ts good">// GOOD: Arguments are explicitly passed to the callback
-const numbers = ['11', '5', '3'].map((n) =&gt; parseInt(n));
+<pre><code class="language-ts good">// GOOD: we only forward the string (and an explicit radix)
+const numbers = ['11', '5', '3'].map((n) =&gt; parseInt(n,10));
```
vapier: missing space after the `,`

tsguide.html
```diff
@@ -1561,8 +1561,8 @@ <h4 id="functions-as-callbacks" class="numbered">Prefer passing arrow functions
 <p>Instead, prefer passing an arrow-function that explicitly forwards parameters to
 the named callback.</p>
 
-<pre><code class="language-ts good">// GOOD: Arguments are explicitly passed to the callback
-const numbers = ['11', '5', '3'].map((n) =&gt; parseInt(n));
+<pre><code class="language-ts good">// GOOD: we only forward the string (and an explicit radix)
```
vapier: I don't think the comment needs to spell things out like this.  the previous comment is fine.

vapier (changes requested):

tsguide.html
KeenoLee: Thank you for your review. I just updated the code based on the comment

tsguide.html
KeenoLee: No problem, I reverted to the previous comment

vapier:
i checked internally and it looks like we also ran into this issue.  unfortunately they switched to using an internal API, so i've asked them to tweak it again to address your feedback while being usable externally.

PR #843, opened 2024-07-19
Update comment from "Good" to "Bad"

dane:
:wave: I know you're not going to merge this PR. It seemed like the easiest way to communicate which line/file had the typo. Cheers!

PR #926, opened 2025-05-16
go: Fix variable name example typo.

brookemitchell:
The first code example under 'Variable name vs. type' is clearly meant to be commented as a 'bad' example.


brookemitchell:
Duplicate, already noted: https://github.com/google/styleguide/pull/843

matttproud:
The fix for this has been committed in the upstream documentation and will propagate into the public style guide documentation on the next export.  Thank you for flagging the issue!

PR #951, opened 2025-11-14
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

It was requested a while back in
- https://github.com/BerriAI/litellm/pull/10232#issuecomment-2826300701

and recommended in
- https://github.com/BerriAI/litellm/issues/5991#issuecomment-2388853171

there indeed could be some false positives but I think it is quite robust -- you can see how few I had to exclude, and some of those could actually be "fixed" to make it cleaner

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,8 @@
+[codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = .git*,*.css,.codespellrc,LICENSE
+check-hidden = true
+# Case sensitive etc. Python regex
+ignore-regex = \b(CLOS|oh noes|xmlns:\S+)\b|github\.com/dsnet/compress/flate|Use "\S+".* not "\S+"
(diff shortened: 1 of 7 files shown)


vapier:
can you attach a full run with these settings against the existing docs ?  or are they all clean other than the fixes you included ?

yarikoptic:
> can you attach a full run with these settings against the existing docs ? or are they all clean other than the fixes you included ?

yes, after those few fixes, it was all clean here as codespell workflow noted in its run:

image

vapier:
sounds great.  we'll look at running codespell internally on these files, and then circle back to enabling this here.

vapier:
i've queued up fixes internally for the various guides and will circle back once they land

PR #965, opened 2026-03-18
go: Export the latest version of the internal guide.

matttproud:
The update revises guidance around line length, import renaming, context usage, sentinel placement, and interface usage.


chressie:
LGTM

Myselfcodingwithzeroknowledge:
It was success but not correct .

PR #980, opened 2026-06-03
Add agent skills support /google-styleguide

rodydavis:
Adds agents skills support for the style guides for various languages. Created as a single skill since most repos have a variety of languages included.

- generated from source
- renders all templates
- adds top level skills directory
- adds python script to keep it always up to date


rodydavis:
Happy to move it to https://github.com/google/skills if needed.

tools/generate_agent_skills.py
```diff
@@ -0,0 +1,903 @@
+"""Google Style Guide Skill Generator.
```
vapier: source code files need proper license & copyright headers

we're using Apache 2.0 for source code in this project

tools/generate_agent_skills.py
```diff
@@ -0,0 +1,903 @@
...
+
+This script automatically regenerates the Markdown style guides and resource
+files from local source files in the repository and external reference pages.
+
+To run this script from the repository root:
+    python3 tools/generate_agent_skills.py
```
vapier: make the file +x and add a proper shebang instead of forcing people to specify `python3` themselves

tools/generate_agent_skills.py
```diff
@@ -0,0 +1,903 @@
...
+"""
+
+from html.parser import HTMLParser
```
vapier: this doesn't follow our style guide ;)

import modules, not classes

tools/generate_agent_skills.py
```diff
@@ -0,0 +1,903 @@
...
+import logging
+import os
+import re
+import shutil
+import subprocess
+import time
+from typing import Dict, List, Optional, Set, Tuple
```
vapier: assume Python 3.11+, so use the standard names in types like `list` instead of `List`

tools/generate_agent_skills.py
```diff
@@ -0,0 +1,903 @@
...
+
+_LINK_REGEX = re.compile(r'\[([^\]]*)\]\(([^)]+)\)')
+
+_STYLEGUIDE_URL_REGEX = re.compile(
+    r'(?<!\()(https?://google\.github\.io/(?:styleguide|swift)'
+    r'[^\s\)\|\"\'<>]*[^\s\)\|\"\'<>\.,])'
+)
+
+# Paths setup
+_SCRIPT_DIR = os.path.dirname(os.path.abspath(__file__))
```
vapier: use Path instead of os.path APIs

tools/generate_agent_skills.py
```diff
@@ -0,0 +1,903 @@
...
+                    else:
+                        syntax_lang = ext_lower.replace('.', '')
+                        if not syntax_lang:
+                            syntax_lang = 'text'
+                        markdown_text = (
+                            f"# {base_name}\n\n"
+                            f"```{syntax_lang}\n{text_content}\n```\n"
+                        )
+                break
+            except Exception as e:
```
vapier: these try blocks are too big, and shouldn't be catching `Exception`.  it seems you want to only cover `subprocess.run` for failing commands.

vapier (changes requested):
if we land it in this repo, we should at least use symlinks instead of duplicating .md files.

somewhere you should document why you need/want to transform the HTML into MD.  maybe in the Python script that does the transformation.

we should add a GH action that runs the script in a "check" or "diff" mode so that it fires on every PR to make sure things stay in sync.

I don't have a strong opinion on putting it in here vs `google/skills`.  if there's guidance in using that project as a single source of Google skills, then that makes sense.

tools/generate_agent_skills.py
rodydavis: I was able to remove the script completely!

rodydavis:
> if we land it in this repo, we should at least use symlinks instead of duplicating .md files.

Done! This worked a lot better and tested installing skills with the symlinks and had no issues.

> somewhere you should document why you need/want to transform the HTML into MD. maybe in the Python script that does the transformation.

No longer doing a transform

> we should add a GH action that runs the script in a "check" or "diff" mode so that it fires on every PR to make sure things stay in sync.

Done and will run on every push

> I don't have a strong opinion on putting it in here vs google/skills. if there's guidance in using that project as a single source of Google skills, then that makes sense.

This repo would be ideal since it could stay in sync

4khg9n7g67-lang (approved):

PR #995, opened 2026-09-15
json: transform xml to html

vapier:
Since browsers don't do this for us anymore, run it through the same publish logic as our other guides to generate the html output.


ccc-hawkins-shenan:
yes yes, of course XSLT is deprecated, because it's the only course i ever took at TriOs College lol.  anyway, that being said, are you sure this is correct?  what exactly is the new link?  i can't seem to find  https://google.github.io/styleguide/jsoncstyleguide.xml  anymore, which would list all the various nodes and the schema etc.  it seems i can now only find a previous version which seems old/deprecated and doesn't give me the same info...

vapier:
"jsoncguide" will be the new name once we update some other stuff, but for now, it should have been "jsoncstyleguide.html" which is linked from the main page.  i'll update the redirect, thx.

